PR #20, opened 2018-05-10
do not compare or send screen widget board_id since it is useless

grosser:
fixes https://github.com/grosser/kennel/pull/19/files

usecase is that a user copy-pasted the api replies and expects there to be no diff
to make future updates not be annoying

### DIFF
--- a/lib/kennel/models/screen.rb
+++ b/lib/kennel/models/screen.rb
@@ -69,6 +69,7 @@ module Kennel
             tile.fetch(:requests).each { |r| r[:conditional_formats] ||= [] }
             tile[:autoscale] = true unless widget[:tile_def].key?(:autoscale)
           end
+          widget.delete :board_id
           widget
         end
       end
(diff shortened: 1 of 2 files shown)


grosser:
@clarkbab does this look alright ?

clarkbab:
@grosser yep I suppose we don't need board_id as it doesn't seem to be very useful. For example, it doesn't seem to matter what value it takes!

One other thing, when there are no widgets for a board we get this error:

`Error:
undefined method `each' for nil:NilClass`

Might want to add a check here for whether the screenboard has any widgets yet: https://github.com/grosser/kennel/blob/master/lib/kennel/models/screen.rb#L55

grosser:
was unable to reproduce that ... screen without widgets works fine and is returned as `[]` when retrieved via api

clarkbab:
@grosser, see the error here where widgets array isn't returned: https://travis-ci.com/zendesk/kennel/builds/73250068?utm_source=github_status&utm_medium=notification.

PR #25, opened 2018-05-30
Add require_full_window option

th11:
wip


grosser:
1.4.0

th11:
@grosser - the best default varies on other monitor settings. I suppose we could parse the query string to determine whether to set the default to true or false.

```
require_full_window a boolean indicating whether this monitor needs a full window of data before it’s evaluated. We highly recommend you set this to False for sparse metrics, otherwise some evaluations are skipped. Default: True for “on average”, “at all times” and “in total” aggregation. False otherwise.
```
https://docs.datadoghq.com/api/?lang=python#create-a-monitor

th11:
I'll give it a shot

PR #26, opened 2018-06-04
Set better defaults for require_full_window option

th11:
For query alerts, set default `require_full_window` option to true for “on average”, “at all times” and “in total” aggregations, otherwise false. This follows DataDogs recommendation (see below). Non query alerts are always set to true which is the current behavior.

> require_full_window a boolean indicating whether this monitor needs a full window of data before it’s evaluated. We highly recommend you set this to False for sparse metrics, otherwise some evaluations are skipped. Default: True for “on average”, “at all times” and “in total” aggregation. False otherwise.

 -- https://docs.datadoghq.com/api/?lang=python#monitors

@grosser

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -26,8 +26,7 @@ module Kennel
         tags: -> { @project.tags },
         multi: ->  { type != "query alert" || query.include?(" by ") },
         critical_recovery: -> { nil },
-        warning_recovery: -> { nil },
-        require_full_window: -> { require_full_window_default(type, query) }
+        warning_recovery: -> { nil }
       )
 
       attr_reader :project
@@ -129,12 +128,10 @@ module Kennel
 
       private
 
-      def require_full_window_default(type, query)
-        return true unless type == "query alert"
+      def require_full_window
         # default 'on_average', 'at_all_times', 'in_total' aggregations to true, otherwise false
         # https://docs.datadoghq.com/ap/#create-a-monitor
-        regex = /\A(avg|min|sum)/
-        query =~ regex ? true : false
+        type != "query alert" || query.start_with?("avg", "min", "sum")
       end
 
       def validate_json(data)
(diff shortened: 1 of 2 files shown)


lib/kennel/models/monitor.rb
```diff
@@ -27,7 +27,7 @@ class Monitor < Base
         multi: ->  { type != "query alert" || query.include?(" by ") },
         critical_recovery: -> { nil },
         warning_recovery: -> { nil },
-        require_full_window: -> { true }
+        require_full_window: -> { require_full_window_default(type, query) }
```
grosser: can just define `require_full_window` and use type + query inside

lib/kennel/models/monitor.rb
```diff
@@ -129,6 +129,14 @@ def diff(actual)
 
       private
 
+      def require_full_window_default(type, query)
+        return true unless type == "query alert"
+        # default 'on_average', 'at_all_times', 'in_total' aggregations to true, otherwise false
+        # https://docs.datadoghq.com/ap/#create-a-monitor
+        regex = /\A(avg|min|sum)/
+        query =~ regex ? true : false
```
grosser: `type != "query alert" && query.start_with?("avg", "min", "sum")`

lib/kennel/models/monitor.rb
th11: Downside to this is that it's not that nice to test, but I can refactor if you'd like.

grosser: yeah it's a bit harder to test, but I'd prefer simple code over simple tests :)

grosser: basically needs a new monitor for each test-case

grosser: `Monitor.new(query: -> { "max" })` etc

th11:
@grosser 💇‍♂️

grosser:
rubocop :D 

FYI https://github.com/grosser/dotfiles/blob/master/bin/pre_commit_rubocop.rb

th11:
whoops - didn't have rubocop install on this version of ruby.

PR #27, opened 2018-06-08
support events hash for dashboard graphs

grosser:


test/kennel/models/dash_test.rb
```diff
@@ -88,6 +88,26 @@ def dash(extra = {})
...
+      ).as_json.must_equal(
+        expected_json.merge(
+          graphs: [
+            {
+              title: "TI",
+              definition: {
+                viz: "V",
+                requests: [{ q: "Q", type: "TY", conditional_formats: [] }],
+                autoscale: true,
+                events: { foo: "bar" }
```
th11: I think this would be an array

grosser:
fixed

grosser:
1.6.0

PR #30, opened 2018-06-20
allow multiple projects to debug their PRs in parallel

grosser:
/fyi @amitizle @waichee @jonmoter @ivathinos @adammw

### DIFF
--- a/template/Readme.md
+++ b/template/Readme.md
@@ -152,7 +152,7 @@ to unblock use the `validate: -> { false }` option.
 ### Debugging locally
 
  - make sure to be on update `master` to not undo other changes
- - run `bundle exec rake kennel:update_datadog`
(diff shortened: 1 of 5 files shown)


template/Readme.md
```diff
@@ -153,6 +153,8 @@ to unblock use the `validate: -> { false }` option.
 
  - make sure to be on update `master` to not undo other changes
  - run `bundle exec rake kennel:update_datadog`
```
waichee: how about  change the command above (line 155) to `PROJECT=truth_service bundle exec rake kennel:update_datadog` instead? since most of the time users are only interested in a single project to test

grosser: yeah good idea, copy-paste instructions should be safe by default :D

amitizle:
That'd be really helpful 👍

Issue #32, opened 2018-06-21
1d interval not supported by Kennel, 24h interval not supported by Datadog


aglover-zendesk:
It looks like Datadog changed some of the interval syntax in their API. When I generate a new monitor via the UI with an evaluation period of 1 day, the query I get in the export is

```
"query": "sum(last_1d):min:...
```

When I execute this in Kennel, Kennel raises an error that my interval is invalid (see https://github.com/grosser/kennel/blob/master/lib/kennel/models/monitor.rb#L158-L163).

If I try to change this to `24h`, Kennel doesn't raise an exception but Datadog chokes on the request:

```
Error 400 during PUT /api/v1/monitor/5315705
request:
...
"query": "sum(last_24h):min:...
...
response:
{"errors":["The value provided for parameter 'query' is invalid"]}
```

Guessing we need to allow `1d` as a valid interval

grosser:
and remove 24h ?

PR #33, opened 2018-06-21
nallowed intervals changed

grosser:
fixes https://github.com/grosser/kennel/issues/32
@aglover-zendesk


aglover-zendesk:
Looks good! I also double checked their [API docs](https://docs.datadoghq.com/api/?lang=python#monitors) to confirm, so this is good to go.

Thanks @grosser!

grosser:
1.8.3

PR #37, opened 2018-07-11
Prevent resources getting filtered from DD when using project filter

henders:
Symptoms are seen when migrating an existing monitor/dashboard into Kennel and then trying to update it from the cmdline.
```
$ PROJECT=doorman bundle exec rake kennel:update_datadog
Generating ... 0.13s
Downloading definitions ... 1.36s
Diffing ... -rake aborted!
Unable to find existing dash with id 52264
```

Problem is that since this dashboard wasn't managed by kennel previously it was missing the tracking kennel id text and so it got filtered out by Kennel.
Now we don't filter if there was no tracking-id.

### DIFF
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -54,11 +54,7 @@ module Kennel
     def calculate_diff
       @update = []
       @delete = []
-
-      actual = Progress.progress "Downloading definitions" do
-        download_definitions
-      end
-      actual.select! { |a| tracking_id(a)&.start_with?("#{@project_filter}:") } if @project_filter
+      actual = filtered_resources_from_dd
 
       Progress.progress "Diffing" do
         details_cache do |cache|
@@ -181,6 +177,19 @@ module Kennel
 
     def tracking_field(a)
       a[:message] ? :message : :description
+    end
+
+    def filtered_resources_from_dd
+      Progress.progress "Downloading definitions" do
+        download_definitions.select do |a|
+          if @project_filter
+            tracking_id = tracking_id(a)
+            tracking_id.nil? || tracking_id.start_with?("#{@project_filter}:")
+          else
+            true
+          end
+        end
+      end
     end
   end
 end
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -54,7 +54,9 @@ module Kennel
     def calculate_diff
       @update = []
       @delete = []
-      actual = filtered_resources_from_dd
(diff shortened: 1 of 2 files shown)


henders:
Rats... hold off on merging this.... It looks like it stopped working... might have broken something when tweaking the tests.

lib/kennel/syncer.rb
```diff
@@ -182,5 +178,18 @@ def tracking_id(a)
     def tracking_field(a)
       a[:message] ? :message : :description
     end
+
+    def filtered_resources_from_dd
+      Progress.progress "Downloading definitions" do
```
grosser: I liked the progress outside ... can make filter a method and pass it `download_definitions ` ?

lib/kennel/syncer.rb
```diff
@@ -182,5 +178,18 @@ def tracking_id(a)
...
+        download_definitions.select do |a|
```
grosser: maybe a `.reject` to not need the else ?

lib/kennel/syncer.rb
henders: Or I can just move the `Progress.progress "Downloading definitions" ...` back to the calling function?

lib/kennel/syncer.rb
henders: actually looks kind of worse:
```
      download_definitions.reject do |a|
        tracking_id = tracking_id(a)
        !tracking_id.nil? && !tracking_id.start_with?("#{@project_filter}:") if @project_filter
      end
```

grosser:
if @project_filter
  definitions = definitions.select { |a|
tracking_id(a)&.start_with?("#{@project_filter}:") }
end

henders:
That puts it back to the original bug... you have to check if the tracking_id is nil before checking if it starts with the relevant project ID.. but I think this will meet your criteria:
```
def filtered_resources_from_dd
    definitions = download_definitions
    if @project_filter
      definitions.select! do |a|
        tracking_id = tracking_id(a)
        tracking_id && !tracking_id.start_with?("#{@project_filter}:") if @project_filter
      end
    end
    definitions
end
```

But I don't think that looks better than what the current one looks like.

henders:
And all good on merging... for some reason the monitor actually got deleted from DD. Is there functionality in Kennel that deletes monitors that are tagged with `Managed by Kennel.....` but don't actually have a Kennel definition?

grosser:
v1.10.1

grosser:
it's hard to know if they had a definition ... so opting for keeping things clean in doubt

PR #39, opened 2018-07-27
validate that only available dimensions are used in is_match

grosser:
`Kennel::Models::Base::ValidationError: foo:bar is_match used with unsupported dimensions ["environment"], allowed dimenions are ["pod"]`

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -138,28 +138,44 @@ module Kennel
       def validate_json(data)
         type = data.fetch(:type)
 
+        # do not allow deprecated type that will be coverted by datadog and then produce a diff
         if type == "metric alert"
           invalid! "type 'metric alert' is deprecated, do not set type to use the default 'query alert'"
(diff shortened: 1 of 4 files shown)


lib/kennel/models/monitor.rb
```diff
@@ -138,30 +138,46 @@ def require_full_window
...
             invalid! "query interval was #{interval}, but must be one of #{QUERY_INTERVALS.join(", ")}"
           end
         end
+
+        if ["query alert", "service check"].include?(type) # TODO: most likely more types need this
+          # verify is_match uses available variables
+          message = data.fetch(:message)
+          used = message.scan(/{{\s*#is_match\s*"([a-zA-Z\d_.-]+).name"/).flatten.uniq
+          allowed = data.fetch(:query)[/by\s*{([^\}]+)}/, 1].to_s.split(/\s*,\s*/)
+          unuspported = used - allowed
```
adammw: typo

PR #44, opened 2018-09-10
Allow setting `evaluation_delay`

khh:
Allow setting `evaluation_delay`. Useful for AWS metrics and recommended by Datadog:

screen shot 2018-09-10 at 13 41 01

@grosser

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -11,12 +11,13 @@ module Kennel
       NON_MULTI_TYPES = ["query alert", "log alert"].freeze # NOTE: event alerts don't seem to return their multi setting
 
       settings(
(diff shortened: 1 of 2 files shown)


khh:
I'm not sure how to fix the integration tests. Any pointers?

grosser:
hmm `+  ~options.evaluation_delay nil -> 0` means that the api did not return evaluation_delay :(
... dug up an old email from datadog saying that `If the evaluation delay is nonzero, it should definitely be returned in all cases, in which case you can then use it to recreate. `
... so I guess we have to covert it to 0 when it is not set

grosser:
so changing `options[:evaluation_delay] ||= nil` -> `options[:evaluation_delay] ||= 0` might do the trick
also comment should change to `# nil / "" / 0 are not returned from the api`

grosser:
I can take over the PR if you want, this is a bit funky, just let me know

grosser:
1.25.0

PR #51, opened 2018-10-03
Project model should add a project tag

smo921:
The tag key name applied by a model's `defaults` should match the model class name.  It makes more sense that the Project model adds a `project` tag.

### DIFF
--- a/lib/kennel/models/project.rb
+++ b/lib/kennel/models/project.rb
@@ -4,7 +4,7 @@ module Kennel
     class Project < Base
       settings :team, :parts, :tags, :slack
       defaults(
-        tags: -> { ["project:#{kennel_id}"] + team.tags },
+        tags: -> { Array(["project:#{kennel_id}"] + team.tags).uniq },
         slack: -> { team.slack }
       )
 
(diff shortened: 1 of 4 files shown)


lib/kennel/models/project.rb
```diff
@@ -4,7 +4,7 @@ module Models
     class Project < Base
       settings :team, :parts, :tags, :slack
       defaults(
-        tags: -> { ["project:#{kennel_id}"] + team.tags },
+        tags: -> { Array(["project:#{kennel_id}"] + team.tags).uniq },
```
smo921: Still need to do this in the subclasses, so this commit is probably unnecessary.

grosser: don't need the `Array`

grosser:
needs a test-fix

yeah this is nice for consistency ... will suck a bit since a few links will break, but naming this service was a mistake :/

smo921:
Completely forgot about this PR and not sure if it is necessary anymore.

grosser:
so still need or not ?

smo921:
I think so.  I hadn't rebased and looked at the code in a while. So I was
unsure if this had already been changed.

PR #52, opened 2018-10-07
add markers

grosser:


lib/kennel/models/dash.rb
```diff
@@ -96,7 +96,9 @@ def render_graphs
 
           requests = Array(queries).map { |q| { q: q, type: type } }
           graph = { title: title, definition: { viz: viz, requests: requests } }
-          graph[:definition][:events] = options[:events] if options[:events]
+          [:events, :markers].each do |key|
```
timbertson-zd: I think this can just be `SUPPORTED_GRAPH_OPTIONS` now?

PR #53, opened 2018-10-11
add importer

grosser:
/cc @zendesk-yoshi-sono @Dagnan @adammw @jonmoter @smo921

```
RESOURCE=dash ID=1234 rake kennel:import
Kennel::Models::Dash.new(
  self,
  title: -> { "Foobar" },
  id: -> { 1234 },
  description: -> { "Foobar" },
  template_variables: -> {
    [
      {
        default: "*",
        prefix: "environment",
        name: "environment"
      },
      {
        default: "*",
        prefix: "pod",
        name: "pod"
      }
    ]
  },
...
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -15,7 +15,7 @@ Keep datadog monitors/dashboards/etc in version control, avoid chaotic managemen
 
  - create a new private `kennel` repo for your organization (do not fork this repo)
  - use the template folder as starting point:
-    ```
+    ```Bash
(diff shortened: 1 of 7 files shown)


Dagnan:
That's pretty cool. From quickly looking at the code I guessed it will use the ruby hash format and not the json straight from Datadog, is that correct?

grosser:
it converts json to ruby hash so it can be copy-pasted into an existing
project

PR #56, opened 2018-11-01
Support resolving IDs for alert graph widgets

alex-stone:
### Description

This follows up https://github.com/grosser/kennel/pull/55, by extending support for resolving IDs to `alert_graph` widgets.

These widgets have a field called `alert_id` that references a monitor ID. Note that this comes as a string when creating via the DataDog UI, but seems to also support it as an integer.

/cc @grosser

### DIFF
--- a/lib/kennel/models/screen.rb
+++ b/lib/kennel/models/screen.rb
@@ -89,6 +89,8 @@ module Kennel
           case widget[:type]
           when "uptime"
             resolve_link(widget, [:monitor, :id], id_map)
+          when "alert_graph"
+            resolve_link(widget, [:alert_id], id_map)
           end
         end
       end
--- a/lib/kennel/models/screen.rb
+++ b/lib/kennel/models/screen.rb
@@ -85,12 +85,11 @@ module Kennel
       end
 
       def resolve_linked_tracking_ids(id_map)
-        uptime_widgets = as_json[:widgets].select { |w| w[:type] == "uptime" && w.dig(:monitor, :id).is_a?(String) }
-        uptime_widgets.each do |uptime_widget|
-          tracking = uptime_widget[:monitor][:id]
-          uptime_widget[:monitor][:id] =
-            id_map[tracking] ||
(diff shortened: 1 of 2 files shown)


lib/kennel/models/screen.rb
```diff
@@ -147,6 +148,37 @@ def widget_defaults(type)
 
         basics.merge(custom)
       end
+
+      def resolve_uptime_widget_link(widget, id_map)
```
grosser: turn this into a single `resolve_link widget, key, type, id_map` and then call `resolve_link widget, [:monitor, :id], Integer, id_map`

alex-stone:
@grosser

## Changes
* Create a single `resolve_link` method that can handle widgets that have different structures for where they store the ID that needs to be resolved.

lib/kennel/models/screen.rb
```diff
@@ -147,6 +148,28 @@ def widget_defaults(type)
...
       end
+
+      def resolve_link(widget, key, id_map)
+        tracking_id = widget.dig(*key)
+
+        return unless link_widget?(tracking_id)
+
+        *id_path, id_key = key
+
+        monitor_path = id_path.empty? ? widget : widget.dig(*id_path)
```
alex-stone: @grosser, thoughts on a cleaner way of handling this?

grosser: nah, as clean as it gets :)

lib/kennel/models/screen.rb
```diff
@@ -147,6 +148,28 @@ def widget_defaults(type)
...
+
+        monitor_path[id_key] =
+          id_map[tracking_id] ||
+          warn("Unable to find #{tracking_id} in existing monitors (they need to be created first to link them)")
+      end
+
+      def link_widget?(tracking_id)
+        tracking_id.is_a?(String) && !string_encoded_id?(tracking_id)
```
grosser: would have gone for `tracking_id.match?(/\A\d+\z/)`

grosser:
1.21.0

grosser:
refactored and updated readme on master

PR #57, opened 2018-11-01
truncate long error messages so something shows up in the comment

grosser:
```
```{"message":"Validation Failed","errors":[{"resource":"CommitComment","code":"custom","field":"body","message":"body is too long (maximum is 65536 characters)"}],"documentation_url":"https://developer.github.com/v3/repos/comments/#create-a-commit-comment"}```
```

@adammw @smedberg


lib/kennel/github_reporter.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 module Kennel
   class GithubReporter
+    MAX_COMMENT_SIZE = 65536
+    TRUNCATED_MSG = "\n```\n... (truncated)" # finish the code block so it look nice
```
adammw: Not sure if this makes sense for errors but anyway, they're unlikely to be that long.

PR #59, opened 2018-11-07
Remove monitors covered by downtime from unmuted alerts

adammw:
/cc @grosser

### DIFF
--- a/lib/kennel/unmuted_alerts.rb
+++ b/lib/kennel/unmuted_alerts.rb
@@ -46,7 +46,7 @@ module Kennel
       def filtered_monitors(api, tag)
         # Download all monitors with given tag
         monitors = Progress.progress("Downloading") do
-          api.list("monitor", monitor_tags: tag, group_states: "all")
+          api.list("monitor", monitor_tags: tag, group_states: "all", with_downtimes: "true")
         end
 
         raise "No monitors for #{tag} found, check your spelling" if monitors.empty?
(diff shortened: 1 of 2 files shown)


lib/kennel/unmuted_alerts.rb
```diff
@@ -66,6 +66,15 @@ def filtered_monitors(api, tag)
           end
         end
 
+        # only keep monitors that are not covered by a downtime
+        monitors.each do |m|
+          next unless m[:matching_downtimes]
+          downtime_groups = m[:matching_downtimes].select { |d| d[:active] }.flat_map { |d| d[:groups] }
+          m[:state][:groups].reject! do |k, _|
+            downtime_groups.include?(k.to_s)
```
grosser: should behave more like line 65
a group only matches a downtime if all the dometimes tags are included in the group

adammw: when i looked at the api response, it looked like it was already expanding the groups array into the groups that would be covered

grosser: ok, looked that up ... groups here is an array of full names (with commas)

test/kennel/unmuted_alerts_test.rb
```diff
@@ -106,6 +106,15 @@ def result
       result.size.must_equal 0
     end
 
+    it "removes monitors that are covered by matching downtimes" do
+      groups = ["pod:pod3", "pod:pod3,project:foo,team:bar"]
+      monitor[:matching_downtimes] = [
+        { end: nil, start: triggered - 10, groups: groups, active: true, scope: ["pod:pod3"] },
+        { end: nil, start: triggered - 10, groups: [], active: true, scope: ["pod:pod7"] }
```
grosser: scope does not matter ?

lib/kennel/unmuted_alerts.rb
grosser: ... if that does not work out because of caching or datadog being slow we can use `scope` and then do the same logic as line 64 with it ...

test/kennel/unmuted_alerts_test.rb
adammw: I don't believe it matters, since the groups end up being expanded to the full group names, but we may need to use it if there's issues with the above...

grosser:
1.22.0

PR #63, opened 2018-12-30
dashes now return nil instead of empty array from api when no variabl…

grosser:
…es are set

complained about that in https://help.datadoghq.com/hc/en-us/requests/189335


grosser:
hmm this seems to only happen for dashes that never had template variables ... so can't merge this for now ...

PR #68, opened 2019-01-24
service check with percent is a thing

grosser:


grosser:
via help.datadoghq.com/hc/requests/193709 "the last() ​ function is for the check alert and the pct_by_status() is for the cluster query"

PR #80, opened 2019-05-17
Add dashboard model

adammw:
cc @grosser

### DIFF
--- a/lib/kennel/importer.rb
+++ b/lib/kennel/importer.rb
@@ -64,8 +64,9 @@ module Kennel
     private
 
     def pretty_print(hash)
-      list = hash.sort_by { |k, _| [SORT_ORDER.index(k) || 999, k] } # important to the front and rest deterministic
-      list.map do |k, v|
+      sort_widgets hash
+
+      sort_hash(hash).map do |k, v|
         pretty_value =
           if v.is_a?(Hash) || (v.is_a?(Array) && !v.all? { |e| e.is_a?(String) })
             # update answer here when changing https://stackoverflow.com/questions/8842546/best-way-to-pretty-print-a-hash
@@ -84,6 +85,20 @@ module Kennel
           end
         "  #{k}: -> {#{pretty_value}}"
       end.join(",\n")
+    end
+
+    # sort dashboard widgets + nesting
+    def sort_widgets(outer)
+      outer[:widgets]&.each do |widgets|
+        definition = widgets[:definition]
+        definition.replace sort_hash(definition)
+        sort_widgets definition
+      end
+    end
+
+    # important to the front and rest deterministic
+    def sort_hash(hash)
+      Hash[hash.sort_by { |k, _| [SORT_ORDER.index(k) || 999, k] }]
     end
   end
 end
--- a/lib/kennel/models/dashboard.rb
+++ b/lib/kennel/models/dashboard.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+module Kennel
+  module Models
+    class Dashboard < Base
+      include TemplateVariables
+      include OptionalValidations
+
+      API_LIST_INCOMPLETE = true
+      WIDGET_DEFAULTS = {
+        time: {},
+        timeframe: "1h"
+      }.freeze
+      SCREEN_DEFAULTS = { template_variables: [] }.freeze
+      WIDGET_READONLY = [:id].freeze
+      READONLY_ATTRIBUTES = Base::READONLY_ATTRIBUTES + [:author_handle, :modified_at, :url]
+
+      settings :id, :title, :description, :widgets, :kennel_id
+
+      defaults(
+        description: -> { "" },
+        widgets: -> { [] },
+        id: -> { nil }
+      )
+
+      attr_reader :project
+
+      def initialize(project, *args)
+        @project = project
+        super(*args)
+      end
+
+      def self.api_resource
+        "dashboard"
+      end
+
+      def as_json
+        return @json if @json
+        @json = {
+          title: "#{title}#{LOCK}",
+          description: description,
+          layout_type: "ordered",
+          is_read_only: false,
+          notify_list: [],
+          widgets: widgets,
+          template_variables: render_template_variables
+        }
+
+        validate_json(@json) if validate
+
+        @json
+      end
+
+      def diff(a)
+        super(a)
+      end
+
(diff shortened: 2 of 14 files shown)


lib/kennel/models/dashboard.rb
```diff
@@ -0,0 +1,110 @@
...
+          widgets: widgets,
+          template_variables: render_template_variables
+        }
+
+        validate_json(@json) if validate
+
+        @json
+      end
+
+      def diff(a)
```
grosser: don't need to declare when not used ?

lib/kennel/models/dashboard.rb
```diff
@@ -0,0 +1,110 @@
...
+        "dashboard"
+      end
+
+      def as_json
+        return @json if @json
+        @json = {
+          title: "#{title}#{LOCK}",
+          description: description,
+          layout_type: "ordered",
+          is_read_only: false,
```
grosser: add to readonly attributes, same as layout_type + notify_list so we never modify it ?

lib/kennel/models/dashboard.rb
```diff
@@ -0,0 +1,110 @@
...
+          *actual[:widgets],
+          *expected[:widgets].flat_map { |w| w.dig(:definition, :widgets) },
+          *actual[:widgets].flat_map { |w| w.dig(:definition, :widgets) }
+        ].compact
+        definitions.each do |definition|
+          definition.delete_if { |key| attributes.include?(key) }
+        end
+      end
+
+      def resolve_link(widget, key, id_map)
```
grosser: a comment on what links are here and how they look like ?

lib/kennel/syncer.rb
```diff
@@ -79,7 +79,7 @@ def calculate_diff
               fill_details(a, cache) if e.class::API_LIST_INCOMPLETE
               diff = e.diff(a)
               @update << [id, e, a, diff] if diff.any?
-            elsif tracking_id(a) # was previously managed
+            elsif tracking_id(a) && !@lookup_map[tracking_id(a)] # was previously managed (need to check lookup map in case of dashboards)
```
grosser: why can't we put the tracking-id into the description ?

lib/kennel/syncer.rb
```diff
@@ -94,7 +94,7 @@ def calculate_diff
     def fill_details(a, cache)
       resource = a.fetch(:api_resource)
       args = [resource, a.fetch(:id)]
-      full = cache.fetch(args, a.fetch(:modified)) do
+      full = cache.fetch(args, a[:modified] || a[:modified_at]) do
```
grosser: great api consistency 🤦‍♂

lib/kennel/syncer.rb
```diff
@@ -146,6 +146,7 @@ def delete_matching_expected(a)
       end
 
       e = @lookup_map["#{a.fetch(:api_resource)}:#{a.fetch(:id)}"] || @lookup_map[tracking_id(a)]
+      return if e && a[:api_resource] != e.class.api_resource
```
grosser: comment on what this does

adammw: so this is the magic that makes it not match the incorrect class type allowing both old Dash and new Dashboard to work. otherwise because the tracking id is the same it would match the wrong object model and fail.

lib/kennel/syncer.rb
adammw: this check is to make sure that not only is the tracking id present, but that it's not present in the lookup map for a different type. this is because "dash" will be present for dashboards, but we don't want to delete that because it will get used/matched later on

grosser:
ok, I think I got everything now ... basic validations for template variables and nested id/styles deletion + nice importers ...

lib/kennel/importer.rb
```diff
@@ -64,8 +64,17 @@ def import(resource, id)
     private
 
     def pretty_print(hash)
-      list = hash.sort_by { |k, _| [SORT_ORDER.index(k) || 999, k] } # important to the front and rest deterministic
-      list.map do |k, v|
+      # sort dashboard widgets + nesting
+      hash[:widgets]&.each do |w1|
```
adammw: extract as sort_widgets method that can call itself?

lib/kennel/models/dashboard.rb
```diff
@@ -0,0 +1,89 @@
...
+      REQUEST_DEFAULTS = {
+        style: { line_width: "normal", palette: "dog_classic", line_type: "solid" }
+      }.freeze
+
+      settings :id, :title, :description, :widgets, :kennel_id, :layout_type
+
+      defaults(
+        description: -> { "" },
+        widgets: -> { [] },
+        id: -> { nil }
```
adammw: default `layout_type`?

lib/kennel/models/dashboard.rb
```diff
@@ -0,0 +1,89 @@
...
+        end
+      end
+
+      def url(id)
+        Utils.path_to_url "/dashboard/#{id}"
+      end
+
+      # expand nested widgets into expected/actual pairs for default resolution
+      # [a, e] -> [[a, e], [aw1, ew1], ...]
+      private_class_method def self.base_pairs(*pair)
```
adammw: should it be separated by a new line? seems unusual, normally i'd expect self methods to be at the top with `class << self`

lib/kennel/models/dashboard.rb
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        @json[:id] = id if id
+
+        validate_json(@json) if validate
+
+        @json
+      end
+
+      def self.normalize(expected, actual)
```
adammw: nit: `class << self`

lib/kennel/models/dashboard.rb
adammw: also why call it base pairs rather than expand_nested_widgets?

lib/kennel/syncer.rb
```diff
@@ -74,8 +74,12 @@ def calculate_diff
         details_cache do |cache|
           actual.each do |a|
             id = a.fetch(:id)
+            e = matching_expected(a)
```
adammw: bitmoji

lib/kennel/models/dashboard.rb
grosser: you got to pick either dash or screen just like the datadog UI :D

Issue #67, opened 2019-01-22
Composite monitors fail the "plan" phase.


zd-carmo:
While creating a composite monitor, it is not possible to run the Plan phase. 
This is because it tries to use the "critical" value. - which is usually not defined in a composite monitor (it uses Boolean logic between other monitors to decide when to trigger each level of alerts).

It should be enough if we define the "Critical" property as nil in case the monitor is a composite one (so it won't need to be explicitly dealt with in the rb files).

grosser:
sounds good, new logic should go into https://github.com/grosser/kennel/blob/735e9f2651135e8ed9f8f1608d90a4d696134f5b/lib/kennel/models/monitor.rb#L71

grosser:
fixed in https://github.com/grosser/kennel/pull/84

PR #83, opened 2019-05-29
Assorted small changes

xaviershay:
You might not want to pull in the whitespace change, since it would cause (harmless) churn on everyone's descriptions.

The composite monitor fix addresses #67.


grosser:
replaced with https://github.com/grosser/kennel/pull/84

PR #85, opened 2019-06-01
new_host_delay

codygill:
adds new_host_delay to be customizable as an option

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -13,6 +13,7 @@ module Kennel
       # defaults that datadog uses when options are not sent, so safe to leave out if our values match their defaults
       MONITOR_OPTION_DEFAULTS = {
         evaluation_delay: nil,
+        new_host_delay: 300,
         timeout_h: 0,
         renotify_interval: 120,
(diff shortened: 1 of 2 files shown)


test/kennel/importer_test.rb
```diff
@@ -189,7 +189,6 @@
           timeout_h: 0,
           include_tags: true,
           no_data_timeframe: nil,
-          new_host_delay: 300,
```
grosser: can't change the api response oO

lib/kennel/models/monitor.rb
```diff
@@ -35,6 +35,7 @@ class Monitor < Base
         notify_no_data: -> { true },
         no_data_timeframe: -> { notify_no_data ? 60 : nil },
         notify_audit: -> { true },
+        new_host_delay: -> { 300 },
```
grosser: should pull from MONITOR_OPTION_DEFAULTS too ? (so we don't have to store the default value everywhere)

grosser:
dammit ... now I need this too ... I'll merge manually

grosser:
v1.37.3

fyi https://github.com/grosser/kennel/commit/424d98a5624eb4435b1510ca6ffc458d92e97749

PR #87, opened 2019-07-02
add model for logging indexes

smo921:
Datadog has recently released the public beta of the API for managing logging indexes.  It would be great to support this in Kennel.

https://docs.datadoghq.com/api/?lang=python#logs-indexes

= TODO
- [ ] Update generator / sync process to include index configuration and index order

### DIFF
--- a/lib/kennel/models/logs_index.rb
+++ b/lib/kennel/models/logs_index.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+# Manage logging index configuration and ordering
+# https://docs.datadoghq.com/api/?lang=python#logs-indexes
+module Kennel
+  module Models
+    class LogsIndex < Base
+      @@index_order = []
+      class << self
+        def sorted
+          @@index_order.reject { |i| i.nil? }
+        end
+
+        def reset!
+          @@index_order = []
+        end
+      end
+
+      settings(:filter, :exclusion_filters, :order)
+      defaults(
+        filter: -> { {query: '*'} },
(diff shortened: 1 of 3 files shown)


lib/kennel/models/logs_index.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# Manage logging index configuration and ordering
+# https://docs.datadoghq.com/api/?lang=python#logs-indexes
+module Kennel
+  module Models
+    class LogsIndex < Base
+      @@index_order = []
```
grosser: not used ?

grosser:
sounds reasonable ... doubt it works in the current state
not a big fan of the `index_order` since it will make testing hard and in general will be brittle/unexpected :(

smo921:
I can ditch the index ordering and use an array, instead of hiding the fact that ordering matters.

grosser:
why do they need to be ordered ?
... use the api like all the other models ?

smo921:
The order of the indexes in the configuration matters because log events are placed into indexes on a first-match based on the index filter.

https://docs.datadoghq.com/api/?lang=bash#update-indexes-order

grosser:
let the users set something like `z-index` so they can adjust the level they want to be in ?
otherwise sort alphabetically ...

smo921:
This is probably all pointless anyway.  If everyone can define a single attribute in their logs to route them to the correct index them the configuration is simple and none of this is necessary.

grosser:
yes, let's start with that :D

PR #93, opened 2019-08-16
task to show monitors with no data

eatwithforks:
@grosser

### DIFF
--- a/lib/kennel/tasks.rb
+++ b/lib/kennel/tasks.rb
@@ -50,6 +50,19 @@ namespace :kennel do
     Kennel::UnmutedAlerts.print(Kennel.send(:api), tag)
   end
 
+  desc "show monitors with no data by TAG, for example TAG=team:foo"
+  task nodata: :environment do
+    tag = ENV["TAG"] || abort("Call with TAG=foo:bar")
+    monitors = Kennel.send(:api).list("monitor", monitor_tags: tag, group_states: "no data")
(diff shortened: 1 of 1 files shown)


lib/kennel/tasks.rb
```diff
@@ -50,6 +50,15 @@
     Kennel::UnmutedAlerts.print(Kennel.send(:api), tag)
   end
 
+  desc "show monitors with no data by TAG, for example TAG=team:foo"
+  task nodata: :environment do
+    monitors = Kennel.send(:api).list("monitor", monitor_tags: tag, group_states: "no data", with_downtimes: "true")
```
grosser: I think this needs to be smarter and say "with each group state being no-data" ... otherwise we get monitors were some things are no-data ?

eatwithforks: updated. did a select! on overall_state: 'no data'

grosser:
so overall state means that all groups are no-data or that some groups are no-data ?

grosser:
... an reduce payload then by using `, group_states: "no-data"` and removing `with_downtimes: "true"`

eatwithforks:
updated to filter by "no data", "no-data" is invalid value. and removed downtimes. the monitors returned are all no data in the overall state

grosser:
... oh boy, tried that ... so many bad monitors :D
.... you know what to do next :trollface:

PR #94, opened 2019-08-20
filter out monitors with nodata:ignore

eatwithforks:
@grosser


lib/kennel/tasks.rb
```diff
@@ -53,8 +53,15 @@
   desc "show monitors with no data by TAG, for example TAG=team:foo"
   task nodata: :environment do
     tag = ENV["TAG"] || abort("Call with TAG=foo:bar")
+    Kennel.out.puts <<~TEXT
```
grosser: - this is not "output", but user interaction -> `Kennel.err`
 - only print when there are problems to not de-sensitize users

PR #97, opened 2019-09-02
Dockerfile into template folder

gastonfournier:
Proposal to add Dockerfile to the template in order to be able to run Kennel from a container


gastonfournier:
Ok, maybe I can create a docker.md explaining how we're using it. 
Indeed the variables needs to be set, we're using .env file to do that so when we mount the current folder in the docker image it just works:
`docker run -it -v $PWD:/usr/src/app {build-id}` 

But I agree it worth explaining a bit how to use it

grosser:
ah that sounds workable, was wondering how you set this up :)

PR #98, opened 2019-09-06
support for auto-linking composite monitors

stewi2:
This adds support for linking composite monitors by id. E.g.

`query: "%{proj:a} || %{proj:b} || %{proj:c}"`

is converted to the corresponding monitor ids if they exist:

`query: 123 || 124 || 125`

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -109,7 +109,7 @@ module Kennel
       end
 
       def resolve_linked_tracking_ids(id_map)
-        if as_json.fetch(:type) == "composite"
+        if as_json[:type] == "composite"
           begin
             as_json[:query].gsub!(/%\{(.*?)\}/) { id_map.fetch($1) }
           rescue KeyError => e
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -108,6 +108,16 @@ module Kennel
         @as_json = data
       end
 
+      def resolve_linked_tracking_ids(id_map)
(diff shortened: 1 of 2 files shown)


lib/kennel/models/monitor.rb
```diff
@@ -108,6 +108,16 @@ def as_json
         @as_json = data
       end
 
+      def resolve_linked_tracking_ids(id_map)
+        if as_json[:type] == "composite"
+          begin
+            as_json[:query].gsub!(/%\{(.*?)\}/) { id_map.fetch($1) }
```
grosser: safer to re-assign with gsub since the query that was passed in might be stored somewhere / be reused

lib/kennel/models/monitor.rb
```diff
@@ -108,6 +108,16 @@ def as_json
...
+          rescue KeyError => e
+            Kennel.err.puts("Unable to find #{e.key} in existing monitors (they need to be created first to link them)")
```
grosser: reuse `resolve_link` (by moving it into base) and don't use fetch, but do `gsub { resolve_link($1) || $1` ... make sure that query is still valid when missing so the initial creation does not get a 400 error

grosser:
I can finish this PR up if you want, let me know :)

lib/kennel/models/monitor.rb
stewi2: Not actually sure what a sensible valid default query would be for a composite monitor, since it seems to require a valid monitor id.

stewi2:
> I can finish this PR up if you want, let me know :)

That would be nice if you don't mind :)

PR #99, opened 2019-09-06
resolve linked monitors in composite monitors

grosser:
@stewi2

replaces https://github.com/grosser/kennel/pull/98

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -109,7 +109,7 @@ module Kennel
       end
 
       def resolve_linked_tracking_ids(id_map)
-        if as_json.fetch(:type) == "composite"
+        if as_json[:type] == "composite"
           begin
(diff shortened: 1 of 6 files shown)


lib/kennel/models/monitor.rb
```diff
@@ -108,6 +108,16 @@ def as_json
         @as_json = data
       end
 
+      # resolve composite monitors ... only works when referenced monitors already exist
+      # since leaving names or bad ids in the query breaks the monitor update
+      def resolve_linked_tracking_ids(id_map)
+        if as_json[:type] == "composite"
```
stewi2: can we be certain that :type will always exist? Should we handle that case here?

grosser: if it's nil then it's not `== "compisite"` 🤷‍♂

stewi2: true... right - I as using fetch() before:)

PR #101, opened 2019-10-05
Add support for scoping alerts

uthark:
To allow further scope alerts.

I.e. it allows to show only alerts from production (assuming monitors using it in grouping)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -175,9 +175,10 @@ To link to existing monitors via their kennel_id
  - figure out project name by converting the class name to snake-case
  - run `PROJECT=foo bundle exec rake kennel:update_datadog` to test changes for a single project
 
-### Listing un-muted alerts
+### Listing unmuted alerts
 
-Run `rake kennel:alerts TAG=service:my-service` to see all un-muted alerts for a given datadog monitor tag.
+Run `rake kennel:alerts TAG=service:my-service SCOPE=scope` to see all un-muted alerts for a given datadog monitor tag.
+Scope is optional and can be omitted.
 
 ### Validating mentions work
 
--- a/lib/kennel/unmuted_alerts.rb
+++ b/lib/kennel/unmuted_alerts.rb
@@ -11,12 +11,16 @@ module Kennel
     }.freeze
 
     class << self
-      def print(api, tag)
+      def print(api, tag, scope)
         monitors = filtered_monitors(api, tag)
         if monitors.empty?
           Kennel.out.puts "No unmuted alerts found"
         else
           monitors.each do |m|
+            if scope
+              m[:state][:groups].select! { |g| g[:name].include?(scope) }
+            end
+            next if m[:state][:groups].empty?
             Kennel.out.puts m[:name]
             Kennel.out.puts Utils.path_to_url("/monitors/#{m[:id]}")
             m[:state][:groups].each do |g|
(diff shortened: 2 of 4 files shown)


uthark:
@grosser Could you please take a look?

grosser:
hmm scope is just a tag again, so how about supporting `TAG=foo,bar` ?

uthark:
It's different. They have monitor_tags which we use. and then we have tags from monitors themselves. Comma didn't work for me.

grosser:
ah yeah .. tags are on the monitor and scopes are on the alerts

lib/kennel/unmuted_alerts.rb
```diff
@@ -11,12 +11,16 @@ class UnmutedAlerts
...
     class << self
-      def print(api, tag)
+      def print(api, tag, scope)
         monitors = filtered_monitors(api, tag)
         if monitors.empty?
           Kennel.out.puts "No unmuted alerts found"
         else
           monitors.each do |m|
+            if scope
+              m[:state][:groups].select! { |g| g[:name].include?(scope) }
```
grosser: `"foo,bar".include?("ba")` would be true, so we should split before testing
```suggestion
              m[:state][:groups].select! { |g| g[:name].split(",").include?(scope) }
```

lib/kennel/unmuted_alerts.rb
```diff
@@ -11,12 +11,16 @@ class UnmutedAlerts
...
+            end
+            next if m[:state][:groups].empty?
```
grosser: should be inside the if ?

lib/kennel/unmuted_alerts.rb
```diff
@@ -44,7 +48,7 @@ def time_since(t)
       end
 
       def filtered_monitors(api, tag)
-        # Download all monitors with given tag
+        # Download all monitors with given tag and scope
```
grosser: has nothing to do with scope
```suggestion
        # Download all monitors with given tag
```

test/kennel/unmuted_alerts_test.rb
```diff
@@ -60,10 +60,33 @@
     it "does not print alerts when there are no monitors" do
       Kennel::UnmutedAlerts.expects(:filtered_monitors).returns([])
       out = Kennel::Utils.capture_stdout do
-        Kennel::UnmutedAlerts.send(:print, nil, tag)
+        Kennel::UnmutedAlerts.send(:print, nil, tag, nil)
       end
       out.must_equal "No unmuted alerts found\n"
     end
+
+    it "filters alerts by scope" do
```
grosser:
```suggestion
    it "only shows groups that match the scope" do
```

Readme.md
```diff
@@ -175,9 +175,10 @@ To link to existing monitors via their kennel_id
  - figure out project name by converting the class name to snake-case
  - run `PROJECT=foo bundle exec rake kennel:update_datadog` to test changes for a single project
 
-### Listing un-muted alerts
+### Listing unmuted alerts
 
-Run `rake kennel:alerts TAG=service:my-service` to see all un-muted alerts for a given datadog monitor tag.
+Run `rake kennel:alerts TAG=service:my-service SCOPE=scope` to see all un-muted alerts for a given datadog monitor tag.
```
grosser:
```suggestion
Run `rake kennel:alerts TAG=service:my-service` to see all un-muted alerts for a given datadog monitor tag.
Optionally filter by a scope with `SCOPE=environment:production`
`
```

grosser:
there is a good chance this will hide alerts that are simply not scoped by a given tag (for example `env:production` but the monitor only has `cluster` tags), so we should be careful with it / not make it the default

PR #105, opened 2019-11-20
validate SLO warning/critical values

smo921:
Add validation of SLO thresholds when the record is created.  It would be cool to have a set of `rules` associated with each model type, and have the validation handled in the `initialize` method in the `Record` class.

Example error from the Datadog API:
```
response:                                                                                                              
{"errors": ["warning must be greater than slo value"]}                                                                 
```

### DIFF
--- a/lib/kennel/models/slo.rb
+++ b/lib/kennel/models/slo.rb
@@ -23,7 +23,10 @@ module Kennel
 
       def initialize(*)
         super
-        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? { |t| t[:warning].to_f <= t[:critical].to_f }
+
+        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? do |t|
+          t.key?(:warning) && !t[:warning].nil? && t[:warning].to_f <= t[:critical].to_f
+        end
       end
 
       def as_json
--- a/lib/kennel/models/slo.rb
+++ b/lib/kennel/models/slo.rb
@@ -19,6 +19,11 @@ module Kennel
         monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) }
       )
 
+      def initialize(project, *args)
+        super(project, *args)
+        raise ValidationError, 'Threshold warning must be less-than critical value' if thresholds.each { |t| t[:warning].to_f < t[:critical].to_f }
+      end
+
       def as_json
         return @as_json if @as_json
         data = {
--- a/lib/kennel/models/slo.rb
+++ b/lib/kennel/models/slo.rb
@@ -21,9 +21,9 @@ module Kennel
         thresholds: -> { DEFAULTS.fetch(:thresholds) }
       )
 
-      def initialize(project, *args)
-        super(project, *args)
-        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? { |t| t[:warning].to_f < t[:critical].to_f }
+      def initialize(*)
+        super
+        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? { |t| t[:warning].to_f <= t[:critical].to_f }
       end
 
       def as_json
--- a/lib/kennel/models/slo.rb
+++ b/lib/kennel/models/slo.rb
@@ -6,7 +6,8 @@ module Kennel
       DEFAULTS = {
         description: nil,
         query: nil,
-        monitor_ids: []
+        monitor_ids: [],
+        thresholds: []
       }.freeze
 
       settings :type, :description, :thresholds, :query, :tags, :monitor_ids, :monitor_tags, :name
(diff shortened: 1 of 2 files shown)


smo921:
:eyes: at test failures . . .

lib/kennel/models/slo.rb
```diff
@@ -19,6 +19,11 @@ class Slo < Record
         monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) }
       )
 
+      def initialize(project, *args)
+        super(project, *args)
+        raise ValidationError, 'Threshold warning must be less-than critical value' if thresholds.each { |t| t[:warning].to_f < t[:critical].to_f }
```
grosser: `.each` is always true ;)

... need `.any?`
... or `thresholds.each {  raise "foo #{warning} #{critical}" if bar`

lib/kennel/models/slo.rb
```diff
@@ -16,9 +17,15 @@ class Slo < Record
         tags: -> { @project.tags },
         query: -> { DEFAULTS.fetch(:query) },
         description: -> { DEFAULTS.fetch(:description) },
-        monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) }
+        monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) },
+        thresholds: -> { DEFAULTS.fetch(:thresholds) }
       )
 
+      def initialize(project, *args)
```
grosser:
```suggestion
      def initialize(*)
```

lib/kennel/models/slo.rb
```diff
@@ -16,9 +17,15 @@ class Slo < Record
...
+        super(project, *args)
```
grosser:
```suggestion
        super
```

lib/kennel/models/slo.rb
```diff
@@ -16,9 +17,15 @@ class Slo < Record
...
+        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? { |t| t[:warning].to_f < t[:critical].to_f }
```
grosser: message does not match check, needs to be `<=`
```suggestion
        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? { |t| t[:warning].to_f <= t[:critical].to_f }
```

test/kennel/models/slo_test.rb
```diff
@@ -44,6 +44,12 @@ def slo(options = {})
     it "stores options" do
       TestSlo.new(project, name: -> { "XXX" }).name.must_equal "XXX"
     end
+
+    it "validates thresholds" do
```
grosser: another test for `warning == threshold`

.gitignore
```diff
@@ -1 +1,2 @@
 template/Gemfile.lock
+tmp/
```
grosser: -> global .gitignore ftw

lib/kennel/models/slo.rb
```diff
@@ -16,9 +17,15 @@ class Slo < Record
...
         query: -> { DEFAULTS.fetch(:query) },
         description: -> { DEFAULTS.fetch(:description) },
-        monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) }
+        monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) },
+        thresholds: -> { DEFAULTS.fetch(:thresholds) }
       )
 
+      def initialize(*)
+        super
+        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? { |t| t[:warning].to_f <= t[:critical].to_f }
```
grosser: that would fail if there is no warning though ? ... are warnings required ?

test/kennel/models/slo_test.rb
```diff
@@ -44,6 +44,15 @@ def slo(options = {})
...
```
grosser: general rule of thumb is 1 assertion per `it`

.gitignore
smo921: TIL

lib/kennel/models/slo.rb
smo921: Trying to find the docs for the SLO Widget api . . . do you have a link?  I can add it to the top of the SLO model file for future reference

smo921: The UI allows the `Warning` to be empty.  I'll address this in another commit.

grosser: yeah good idea ... can also verify via the UI if it lets you not enter a warning level

lib/kennel/models/slo.rb
```diff
@@ -16,9 +17,18 @@ class Slo < Record
...
-        monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) }
+        monitor_ids: -> { DEFAULTS.fetch(:monitor_ids) },
+        thresholds: -> { DEFAULTS.fetch(:thresholds) }
       )
 
+      def initialize(*)
+        super
+
+        raise ValidationError, "Threshold warning must be greater-than critical value" if thresholds.any? do |t|
+          t.key?(:warning) && !t[:warning].nil? && t[:warning].to_f <= t[:critical].to_f
```
grosser:
```suggestion
          t[:warning] && t[:warning].to_f <= t[:critical].to_f
```

grosser: and then best make this a `if foo -newline-` multiline trailing ifs are kinda funky

test/kennel/models/slo_test.rb
```diff
@@ -44,6 +44,24 @@ def slo(options = {})
...
+        Kennel::Models::Slo.new(project, thresholds: -> { [{ critical: 99 }] })
+      end
+
+      it "is invalid if warning > critical" do
+        assert_raises Kennel::Models::Record::ValidationError do
+          Kennel::Models::Slo.new(project, thresholds: -> { [{ warning: 0, critical: 99 }] })
+        end
+      end
+
+      it "is invalid warning != critical" do
```
grosser:
```suggestion
      it "is invalid warning == critical" do
```

grosser: is that actually true ? / does the UI complain ?
... if not then we should not impose this, as `kennel` is meant to be a transparent layer over datadog and not opinionated

smo921: Screen Shot 2019-11-21 at 8 39 04 AM

PR #112, opened 2020-02-26
Allow 2d and 1w intervals for anomaly monitors

gabetax:
Datadog's anomaly monitor UI implicitly creates queries like this:

```
avg(last_1w):anomalies(sum:foo.domain_event{env:production} by {event}.as_count(), 'agile', 5, direction='below', alert_window='last_1d', interval=3600, count_default_zero='true', seasonality='weekly') >= 1
```

The value in the first field is not present in the UI, but scales ahead of the "alert window" you define.

Using the above query definition prior to this change raises this exception:

```
Kennel::ValidationError: foo query interval was 1w, but must be one of 1m, 5m, 10m, 15m, 30m, 1h, 2h, 4h, 1d
```

I've confirmed this change allows the above query to generate successfully in the downstream monitor repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 template/Gemfile.lock
+tmp/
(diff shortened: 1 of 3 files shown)


grosser:
1.66.2

PR #119, opened 2020-06-25
set default template_variable_presets to nil

uthark:


lib/kennel/models/dashboard.rb
```diff
@@ -103,10 +103,10 @@ def as_json
           title: "#{title}#{LOCK}",
           description: description,
           template_variables: render_template_variables,
-          template_variable_presets: template_variable_presets,
           widgets: all_widgets
         }
 
+        @json[:template_variable_presets] = template_variable_presets if template_variable_presets
```
grosser, line 109: we need to set the nil because otherwise if you remove presets then it would not deleted

grosser:
oh boy ... `+  +template_variable_presets nil -> nil`
... so I think I was wrong and it's not that it's `nil` but that it is not returned at all :(

grosser:
needs ignore_defaults I think ... can steal from monitor.rb

uthark:
@grosser PTAL. integration tests are now passing too.

grosser:
1.72.2

PR #122, opened 2020-09-03
Make slo regex more forgiving on different url format

nguyenmv2:
### Description

It's possible for the SLO URL in Datadog to contain other query strings when other filtering criteria are applied (ex: `https://foo.datadoghq.com/slo?query="bar"&slo_id=<slo_id>`). 
Currently the regex for SLO assumes that the `slo_id` is the first query string param.

- Allow characters to appear before the `slo_id` query param.


grosser:
1.76.2

PR #128, opened 2021-03-08
Remove query alert interval verification

mbigornia:
Aside from the normal interval such as 5 minutes, 15 minutes, 1 hour, etc. query interval selection now includes an option to set a custom value between 1 minute to 48 hours. This means something like this should be valid:

`max(last_91m):min:datadog.agent.running{hostgroup:jumpbox} by {host,pod}.fill(zero) < 1`

https://docs.datadoghq.com/monitors/monitor_types/metric/?tab=threshold


grosser:
thx!

grosser:
1.84.0

PR #130, opened 2021-05-28
Tracking refactor

adammw:
Refactor tracking id info into record model, allowing for model to override the logic later.

- [X] benchmark slowdown compared to master


grosser:
FYI 
- will most likely need to extract the tracking-id -> project:kennel_id and back conversion
- will run into trouble with kennel_ids atm allowing all kinds of funkyness like whitespace etc ... so limiting that first might be good

grosser:
FYI 
- will most likely need to extract the tracking-id -> project:kennel_id and back conversion
- will run into trouble with kennel_ids atm allowing all kinds of funkyness like whitespace etc ... so limiting that first might be good

PR #131, opened 2021-06-09
Add LogMetric record and APIs

adammw:
Add log metric record for creating custom metrics from log queries.
https://docs.datadoghq.com/api/latest/logs-metrics/

Note that the user's corresponding to the API/Application key must have the "Logs Public Config API" permissions to hit these APIs, and is separate from permissions to edit custom log metrics via the Datadog UI.

TODO:
- [ ] Changing metric isn't being detected correctly and shows "nothing to do"
- [ ] Check everywhere else TRACKING_FIELD is used, and make sure it'll be compatible
- [ ] Write/update tests

### DIFF
--- a/lib/kennel/api.rb
+++ b/lib/kennel/api.rb
@@ -11,8 +11,8 @@ module Kennel
     end
 
     def show(api_resource, id, params = {})
-      reply = request :get, "/api/v1/#{api_resource}/#{id}", params: params
-      api_resource == "slo" ? reply[:data] : reply
+      reply = request :get, api_url(api_resource, id), params: params
+      api_resource == "slo" || api_resource == 'logs/config/metrics' ? reply[:data] : reply
     end
 
     def list(api_resource, params = {})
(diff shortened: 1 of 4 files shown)


lib/kennel/api.rb
```diff
@@ -23,33 +23,42 @@ def list(api_resource, params = {})
...
 
     def create(api_resource, attributes)
-      reply = request :post, "/api/v1/#{api_resource}", body: attributes
-      api_resource == "slo" ? reply[:data].first : reply
+      attributes = { data: attributes } if api_resource == 'logs/config/metrics'
+      reply = request :post, api_url(api_resource), body: attributes
+      if api_resource == "slo"
+        reply[:data].first
+      elsif api_resource == 'logs/config/metrics'
+        reply[:data]
```
grosser, line 45: if we make the model generate `data: attributes` then we don't need to pack/unpack here
... but not much different either way

adammw: i tried that but then it shows a permadiff when fetching for updates, since i think it already checks within :data for list?

lib/kennel/models/log_metric.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+module Kennel
+  module Models
+    class LogMetric < Record
+      VIRTUAL_ID = true
+
+      settings(:metric, :query, :group_by, :aggregation_type, :aggregation_path)
```
grosser, line 8: why use metric when `id` already exists ?

lib/kennel/models/log_metric.rb
```diff
@@ -0,0 +1,76 @@
...
+      end
+
+      def as_json
+        return @as_json if @as_json
+        data = {
+          id: metric,
+          type: 'logs_metrics',
+          attributes: {
+            filter: {
+              query: query.dup # unfreeze
```
grosser, line 51: say why we need the dup

lib/kennel/models/log_metric.rb
```diff
@@ -0,0 +1,76 @@
...
+        @as_json = data
+      end
+
+      def add_tracking_id
+        json = as_json
+        if self.class.parse_tracking_id(json)
+          invalid! "remove \"-managed_by_kennel\" portion from query to copy a resource"
+        end
+        json.dig(*TRACKING_FIELD).replace(json.dig(*TRACKING_FIELD) + " -managed_by_kennel:\"#{tracking_id}\"")
+      end
```
grosser, line 73: add some todo that this should be unified with record

lib/kennel/syncer.rb
```diff
@@ -153,7 +153,7 @@ def ensure_all_ids_found
       @expected.each do |e|
         next unless id = e.id
         resource = e.class.api_resource
-        raise "Unable to find existing #{resource} with id #{id}\nIf the #{resource} was deleted, remove the `id: -> { #{e.id} }` line."
+        raise "Unable to find existing #{resource} with id #{id}\nIf the #{resource} was deleted, remove the `id: -> { #{e.id.inspect} }` line." unless e.class::VIRTUAL_ID
```
grosser, line 156: need to set VIRTUAL_ID false in record ?
... also can you elaborate why this is necessary

grosser:
this will break all users that don't have that permission then :(
... unless there is a nice way to opt everyone in
... might mean the resource needs to be opt-in but that would be a big rabbithole :(

PR #139, opened 2021-11-04
Synthetics as monitors

zdrve:
There are several commits in this PR, which don't all _need_ to be here.

The key change is:
 * the keys in id_map need to include the target object type (so that "synthetics/tests foo:bar" and "monitor foo:bar" can resolve to different IDs)
 * This patch:

```ruby
        if a[:klass].api_resource == "synthetics/tests"
          @id_map.add(Kennel::Models::Monitor.api_resource, tracking_id, a.fetch(:monitor_id))
        end
```

However, this PR includes more than that. In particular:

 * moving id_map from a plain Hash to the new `IdMap` class, so the isolation is cleaner
 * introducing the "resolve" method, to simplify the call sites
 * using more accurate test data (SLO IDs are strings not integers)
 * removing one apparently-invalid test


lib/kennel/syncer.rb
```diff
@@ -254,11 +254,17 @@ def prevent_irreversible_partial_updates
     end
 
     def populate_id_map(expected, actual)
+      expected.each do |e|
+        @id_map.add_new(e.class.api_resource, e.tracking_id)
+      end
```
grosser, line 259: this seems wrong, not everything that is "expected" is new

lib/kennel/id_map.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+module Kennel
+  class IdMap
+    NEW = :new
+
+    def initialize
+      @map = Hash.new { |h, k| h[k] = {} }
+    end
+
+    def add(type, tracking_id, id)
```
grosser, line 10: more of a `set` ?

lib/kennel/id_map.rb
```diff
@@ -0,0 +1,26 @@
...
+      @map[type][tracking_id] = id
+    end
+
+    def add_new(type, tracking_id)
+      @map[type][tracking_id] = NEW
+    end
```
grosser, line 16: not a big fan of this interface since it would require 2 new methods for each thing (like NEW), better to set `IdMap::NEW` and compare against `IdMap::NEW` ?

lib/kennel/id_map.rb
```diff
@@ -0,0 +1,26 @@
...
+module Kennel
+  class IdMap
+    NEW = :new
+
+    def initialize
+      @map = Hash.new { |h, k| h[k] = {} }
+    end
+
+    def add(type, tracking_id, id)
+      @map[type][tracking_id] = id
```
grosser, line 11: FYI this would also work, but would create more bogus array objects
```suggestion
      @map[[type, tracking_id]] = id
```

lib/kennel/syncer.rb
zdrve: True, but it's immediately overwritten by the next section.

grosser: ah true, the old code did the sane but more complicated 👍

grosser:
lucky v1.100.0 ;)
... need to wait for my other bump PR to get merged or some validations will fail

PR #140, opened 2021-11-04
Add padlock icon to synthetic test names

zdrve:
Like we do for the other record types.


grosser (approved):
good catch, thx! :)

grosser:
1.98.3

PR #141, opened 2021-11-05
json nodata report

grosser:
@zdrve


zdrve:
You've presumably noticed the almost-duplicate code to this in zendesk/kennel, right?

grosser:
this is it's replacement :)

PR #144, opened 2021-11-15
Syncer bugs

zdrve:
Two bug fixes:

 * when resolving a monitor ID using a synthetic, that should also work for synthetics being created during this run
 * when using a project filter, it should be possible to refer to existing items outside of the filter

### DIFF
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -263,7 +263,8 @@ module Kennel
 
       actual.each do |a|
         next unless tracking_id = a.fetch(:tracking_id)
-        next unless @id_map.get(a.fetch(:klass).api_resource, tracking_id)
+        next unless @id_map.get(a.fetch(:klass).api_resource, tracking_id) \
+          || (@project_filter && !tracking_id.start_with?("#{@project_filter}:"))
 
         @id_map.set(a.fetch(:klass).api_resource, tracking_id, a.fetch(:id))
         if a[:klass].api_resource == "synthetics/tests"
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -256,6 +256,9 @@ module Kennel
     def populate_id_map(expected, actual)
       expected.each do |e|
         @id_map.set_new(e.class.api_resource, e.tracking_id)
+        if e.class.api_resource == "synthetics/tests"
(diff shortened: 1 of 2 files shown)


lib/kennel/syncer.rb
```diff
@@ -256,11 +256,15 @@ def prevent_irreversible_partial_updates
...
+        if e.class.api_resource == "synthetics/tests"
+          @id_map.set_new(Kennel::Models::Monitor.api_resource, e.tracking_id)
+        end
       end
 
       actual.each do |a|
         next unless tracking_id = a.fetch(:tracking_id)
-        next unless @id_map.get(a.fetch(:klass).api_resource, tracking_id)
+        next unless @id_map.get(a.fetch(:klass).api_resource, tracking_id) \
+          || (@project_filter && !tracking_id.start_with?("#{@project_filter}:"))
```
grosser, line 267: should we just remove this and always set it ?

grosser: the actual is the source of truth after all 🤷

zdrve:
```
  1) Failure:
Kennel::Syncer::#update#test_0009_fails on to-be-deleted dependencies [/Users/revans/git/github.com/grosser/kennel/test/kennel/syncer_test.rb:557]:
unexpected invocation: #<Mock:Api>.create("slo", {:name => "x🔒", :description => "x\n-- Managed by kennel a:c in test/test_helper.rb, do not modify manually", :thresholds => [], :monitor_ids => [456], :tags => ["service:a", "team:test_team"], :type => "metric"})
```

If we were to remove this line then we'd either need to find another fix, or consciously not care about this behaviour.

zdrve:
Is there anything else needed from me before this can be merged?

grosser:
taking another look ... that line is still sus to me

lib/kennel/syncer.rb
zdrve: "Only use this actual (i.e. make it available for resolution) if it either corresponds to an expected, or ( we're using a project filter and it's from a different project )"

zdrve: > the actual is the source of truth after all 🤷

Not when you're using a project filter it isn't.

grosser: ok, get it now ... not perfect but good enough ...

grosser:
1.102.1

PR #148, opened 2021-11-29
clarify warning

grosser:


zdrve:
FWIW I didn't understand the message before and I still don't. What's it trying to achieve?

grosser:
this was to prevent branch updates from breaking master
but it can go now since it's only a warning on master ...

PR #147, opened 2021-11-20
Turn missing imports into a warning, if running master ci

zdrve:
(These examples force a failure by changing both an `id` and `kennel_id` from a clean codebase):

Non-master CI:

```
harvey:kennel revans$ with-ruby - bundle exec rake generate kennel:ci
Generating ... 2.84s
Storing ... 4.39s
Downloading definitions ... 8.55s
Diffing ... -rake aborted!
Unable to find existing dashboard with id a33-zbi-zzz
If the dashboard was deleted, remove the `id: -> { a33-zbi-zzz }` line.
```

Master CI:

```
harvey:kennel revans$ GITHUB_REF=refs/heads/master GITHUB_EVENT_NAME=push  with-ruby - bundle exec rake generate kennel:ci
Generating ... 3.13s
Storing ... 4.58s
Downloading definitions ... 8.9s
Diffing ... 15.08s
Plan:
Warning: dashboard admin_centre:admin-center-errors-dashboard2 specifies id a33-zbi-zzz, but no such dashboard exists. 'id' will be ignored. Remove the `id: -> { a33-zbi-zzz }` line.
Create dashboard admin_centre:admin-center-errors-dashboard2
Delete dashboard admin_centre:admin-center-errors-dashboard
Execute Plan ? -  press 'y' to continue: ^C
```

Note that this introduces one extra line of code in an area that already lacks test coverage.

PR #149, opened 2021-11-29
remove deletion warning because deletion now only warns on master

grosser:
after https://github.com/grosser/kennel/pull/147


grosser:
@zdrve

zdrve (approved):

PR #150, opened 2022-03-21
Refuse update of dashboard layout_type

zdrve:
A sketch. What do you think?

We get post-merge failures like this one https://github.com/zendesk/kennel/runs/5622881439?check_suite_focus=true : 

```json
{"errors": ["'layout_type' is not editable for existing dashboards. This dashboard uses ordered layout."]}
```

It could be good to trap those before merge. This is one way of doing that.

But: how far do we want to go down the path of reproducing datadog's validations in kennel?

### DIFF
--- a/lib/kennel/models/dashboard.rb
+++ b/lib/kennel/models/dashboard.rb
@@ -205,9 +205,9 @@ module Kennel
         end
       end
 
-      def validate_update!(actual, diffs)
-        if diff = diffs.find { |diff| diff[1] == "layout_type" }
-          raise "Datadog does not allow update of #{diff[1]} (in #{tracking_id}, #{diff[2].inspect} -> #{diff[3].inspect})"
+      def validate_update!(_actual, diffs)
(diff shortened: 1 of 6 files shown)


zdrve:
1. What about the monitor types? (It's the "type" field or something that can't be changed, right? Rings a bell...)
2. How about meeting half-way: get rid of the "delete" hook, since (a) we have no use case for it yet, and it's hard to envisage what one might be and (b) it operates rather messily anyway, because it gets a plain hash instead of a Record. tl;dr: keep create+update; ditch the delete hook.

grosser:
- idk what the usecase for the create hook is, so prefer to drop it
- yes monitor type should also be enforced

zdrve:
I'm not sure what's going on with the checks on this PR. Could you either merge this, or tell me what I need to do to get this closer to being merged? Thanks!

grosser:
CI being weird ... rebased this PR and then checked CI manually and it was green 🤷

PR #152, opened 2022-03-24
simplify validate_update

grosser:
@zdrve

### DIFF
--- a/lib/kennel/models/dashboard.rb
+++ b/lib/kennel/models/dashboard.rb
@@ -209,9 +209,9 @@ module Kennel
         end
       end
 
-      def validate_update!(_actual, diffs)
+      def validate_update!(_actuals, diffs)
         if bad_diff = diffs.find { |diff| diff[1] == "layout_type" }
-          raise "Datadog does not allow update of #{bad_diff[1]} (in #{tracking_id}, #{bad_diff[2].inspect} -> #{bad_diff[3].inspect})"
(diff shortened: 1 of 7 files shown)


lib/kennel/models/dashboard.rb
```diff
@@ -209,9 +209,9 @@ def resolve_linked_tracking_ids!(id_map, **args)
         end
       end
 
-      def validate_update!(_actual, diffs)
+      def validate_update!(diffs)
         if bad_diff = diffs.find { |diff| diff[1] == "layout_type" }
-          raise "Datadog does not allow update of #{bad_diff[1]} (in #{tracking_id}, #{bad_diff[2].inspect} -> #{bad_diff[3].inspect})"
+          invalid! "Datadog does not allow update of #{bad_diff[1]} (#{bad_diff[2].inspect} -> #{bad_diff[3].inspect})"
```
zdrve, line 214: 👍

lib/kennel/syncer.rb
```diff
@@ -214,8 +214,8 @@ def print_diff(diff)
     # We've already validated the desired objects ('generated') in isolation.
     # Now that we have made the plan, we can perform some more validation.
     def validate_plan
-      @update.each do |_, expected, actual, diffs|
-        expected.validate_update!(actual, diffs)
+      @update.each do |_, expected, _, diffs|
+        expected.validate_update!(diffs)
```
zdrve, line 218: This feels to me like tailoring a bit too much to the _current_ use case. Providing the `actual` could be useful to the receiver and costs us almost nothing to provide.

grosser: hmm true, might be useful

PR #153, opened 2022-04-07
dont alert on no data during renotify if "no data" option is selected

waichee:
Currently if renotify option is selected, the monitor will renotify on both alert and no data. 
We should only do renotify on alert if `notify_no_data` option is not selected.

How the monitor json looks like
```
{
	"name": "som name",
	"type": "query alert",
	"query": xxx
	"options": {
		"thresholds": {
			"critical": 80
		},
		"notify_audit": false,
		"require_full_window": false,
		"notify_no_data": false,
		"renotify_interval": 30,
		"silenced": {},
		"include_tags": true,
		"new_host_delay": 300,
		"escalation_message": "",
		"renotify_statuses": [
			"alert",
			"no data",
			"warn"
		]
	},
	"priority": null,
	"restricted_roles": null
}
```

@grosser

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -76,8 +76,9 @@ module Kennel
             escalation_message: Utils.presence(escalation_message.strip),
             evaluation_delay: evaluation_delay,
             locked: false, # setting this to true prevents any edit and breaks updates when using replace workflow
-            renotify_interval: renotify_interval || 0
-          }
+            renotify_interval: renotify_interval || 0,
+            renotify_statuses: notify_no_data ? ["alert", "no data"] : ["alert"]
+        }
         }
 
         data[:id] = id if id
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -81,12 +81,6 @@ module Kennel
         }
 
         data[:id] = id if id
-        if renotify_interval != 0
-          statuses = ["alert"]
-          statuses << "no data" if notify_no_data
-          statuses << "warn" if warning
-          data[:options][:renotify_statuses] = statuses
-        end
 
         options = data[:options]
         if data.fetch(:type) != "composite"
@@ -116,6 +110,14 @@ module Kennel
 
         # Datadog requires only either new_group_delay or new_host_delay, never both
         options.delete(options[:new_group_delay] ? :new_host_delay : :new_group_delay)
+
+        # Add in statuses where we would re notify on. Possible values: alert, no data, warn
+        if options[:renotify_interval] != 0
+          statuses = ["alert"]
+          statuses << "no data" if options[:notify_no_data]
+          statuses << "warn" if options.dig(:thresholds, :warning)
+          options[:renotify_statuses] = statuses
+        end
 
         validate_json(data) if validate
 
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -76,12 +76,15 @@ module Kennel
             escalation_message: Utils.presence(escalation_message.strip),
             evaluation_delay: evaluation_delay,
             locked: false, # setting this to true prevents any edit and breaks updates when using replace workflow
-            renotify_interval: renotify_interval || 0,
-            renotify_statuses: notify_no_data ? ["alert", "no data"] : ["alert"]
-        }
+            renotify_interval: renotify_interval || 0
+          }
         }
 
         data[:id] = id if id
+
+        if renotify_interval != 0 && !notify_no_data
+          data[:options][:renotify_statuses] =["alert"]
+        end
 
         options = data[:options]
         if data.fetch(:type) != "composite"
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -81,9 +81,11 @@ module Kennel
         }
 
         data[:id] = id if id
-
-        if renotify_interval != 0 && !notify_no_data
-          data[:options][:renotify_statuses] =["alert"]
+        if renotify_interval != 0
+          statuses = ["alert"]
+          statuses << "no data" if notify_no_data
+          statuses << "warn" if warning
+          data[:options][:renotify_statuses] = statuses
         end
 
         options = data[:options]
--- a/test/kennel/models/monitor_test.rb
+++ b/test/kennel/models/monitor_test.rb
@@ -41,10 +41,6 @@ describe Kennel::Models::Monitor do
         evaluation_delay: nil,
         locked: false,
         renotify_interval: 0,
-        renotify_statuses: [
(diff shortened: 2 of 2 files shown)


lib/kennel/models/monitor.rb
```diff
@@ -76,8 +76,9 @@ def as_json
             escalation_message: Utils.presence(escalation_message.strip),
             evaluation_delay: evaluation_delay,
             locked: false, # setting this to true prevents any edit and breaks updates when using replace workflow
-            renotify_interval: renotify_interval || 0
-          }
+            renotify_interval: renotify_interval || 0,
+            renotify_statuses: notify_no_data ? ["alert", "no data"] : ["alert"]
```
grosser, line 80: can we only set renotify_statuses if renotify_interval is not 0 ?
... I'm hoping that would produce less diff in the `rake generate` and get the same result
also it should be include warning when there is a warning configured ?

waichee: sure ill change it to renotify_interval.

>  also it should be include warning when there is a warning configured ?
what does it mean?

mesge: I think the alert state can be `warn`, `alert`, or `no data` so if they've got a warning alert threshold configured then renotify statuses might wanna be `["alert", "warn", "no data"]`

waichee: ah i see. thanks ill add warn

waichee: 💇  @grosser

grosser:
ran generate and got
```
     "renotify_interval": 0,
     "renotify_statuses": [
       "alert",
       "warn"
     ],
```

lib/kennel/models/monitor.rb
```diff
@@ -81,6 +81,12 @@ def as_json
         }
 
         data[:id] = id if id
+        if renotify_interval != 0
```
grosser, line 84:
```suggestion
        if data[:options][:renotify_interval] != 0
```

test/kennel/models/monitor_test.rb
```diff
@@ -155,6 +155,36 @@ def monitor(options = {})
...
+        notify_no_data: -> { false },
+        ).as_json[:options][:renotify_statuses].must_equal ["alert"]
+    end
+
+    it "sets renotify_statuses to alert and warn when renotify_interval and warning are set" do
+      monitor(
+        renotify_interval: -> { 10 },
+        notify_no_data: -> { false },
+        warning: -> { 10 },
+        ).as_json[:options][:renotify_statuses].must_equal ["alert", "warn"]
```
grosser, line 170: `rubocop -a test/kennel/models/monitor_test.rb`
can also do as part of commit with https://github.com/grosser/dotfiles/blob/master/bin/pre_commit_rubocop.rb

lib/kennel/models/monitor.rb
```diff
@@ -81,6 +81,12 @@ def as_json
...
+          statuses = ["alert"]
+          statuses << "no data" if notify_no_data
+          statuses << "warn" if warning
```
grosser, line 87: best to use the generated has for these too, to avoid duplicate method calls (for example calling `warning` twice)
so basically move this whole block just above the `validate_json` call

waichee:
i tested with `rake plan/generate` . Verified that it didnt add extra renotify statuses for monitor without it, e.g. f[ront end event ingestor http API 0 monitor](https://github.com/zendesk/kennel/blob/2b3dbc74fa3223688bc202b6747216a512e9510e/projects/sunshine/api.rb#L78)

```
$ g status | head -20                  [2.6.5]
On branch wyau/update-add-alert
Your branch is up to date with 'origin/wyau/update-add-alert'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Gemfile
	modified:   Gemfile.lock
	modified:   generated/account_data_deletion/auaora-cpu-pid-sensor.json
	modified:   generated/account_data_deletion/auaora-maxwell-filters-lag-pid-sensor.json
	modified:   generated/account_data_deletion/auaora-trx-pid-sensor.json
	modified:   generated/account_data_deletion/data-deletion-audits-failed-persistently.json
	modified:   generated/answer_bot_flow_director/scale-limits.json
	modified:   generated/auto_translation/translation-reaching-quota-limit.json
	modified:   generated/billing_app/404-when-loading-assets-for-billing.json
	modified:   generated/billing_manatees_project/billing-non-spp-no-legacy-subscriptions-have-been-created-in-the-last-18-hours.json
	modified:   generated/billing_manatees_project/billing-spp-no-subscriptions-have-been-created-in-the-last-4-hours.json
	modified:   generated/cassandra/cassandra-disk-usage-on-node-is-80.json
	modified:   generated/cassandra/cassandra-global-read-latency.json
	modified:   generated/cassandra/cassandra-global-write-latency.json
$ g status | grep frontend-event-ingestor-overall-api-request-rate-
```

Though it looks like its still changing about 560 files with this rather than 800++ files. 

Added a unit test for nil renotify interval too.

lib/kennel/models/monitor.rb
```diff
@@ -38,20 +38,20 @@ class Monitor < Record
         message: -> { "\n\n#{project.mention}" },
         escalation_message: -> { DEFAULT_ESCALATION_MESSAGE.first },
         renotify_interval: -> { project.team.renotify_interval },
-        warning: -> { nil },
```
waichee, line 41: rubocop didnt like `{ nil }`

grosser: seems weird since the rubocop version is hardcoded ... so maybe you ran it with a different version of rubocop or no .rubocop.yml ?

grosser: .... also this exists in multiple files, so please restore. If we decide to change that it should be it's own PR.

waichee: ya i did run directly `rubocop` without `bundle exec` at the front. i can put back the nil

test/kennel/models/monitor_test.rb
```diff
@@ -155,6 +155,41 @@ def monitor(options = {})
       ).as_json[:options][:no_data_timeframe].must_be_nil
     end
 
+    it "sets renotify_statuses to alert only when renotify_interval is greater than 1, and notify no data and warning are disabled" do
```
grosser, line 158: `describe "renotify_statuses" do` might help keep these lines shorter

grosser:
I'll merge once CI is happy thanks! :)

grosser:
1.109.0

test/kennel/models/monitor_test.rb
waichee: sure updated

waichee:
thanks @grosser

PR #154, opened 2022-04-14
Multi project filter

zdrve:
Context: https://zendesk.slack.com/archives/C81T33XTK/p1649800127391659?thread_ts=1649795608.256769&cid=C81T33XTK

I've found myself wanting this feature every now and then.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -248,6 +248,7 @@ so they can be created in a single update and can be re-created if any of them i
  - rebase on updated `master` to not undo other changes
  - figure out project name by converting the class name to snake-case
  - run `PROJECT=foo bundle exec rake kennel:update_datadog` to test changes for a single project (monitors: remove mentions while debugging to avoid alert spam)
+   - you can also use `PROJECT=foo:bar:...` if your work spans multiple projects
 
 ### Reuse
 Add to `parts/<folder>`.
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -63,19 +63,21 @@ module Kennel
 
     def store(parts)
       Progress.progress "Storing" do
-        old = Dir["generated/#{project_filter || "**"}/*"]
-        used = []
+        (project_filter || ["**"]).each do |project_pattern|
+          old = Dir["generated/#{project_pattern}/*"]
+          used = []
 
-        Utils.parallel(parts, max: 2) do |part|
-          path = "generated/#{part.tracking_id.tr("/", ":").sub(":", "/")}.json"
-          used << File.dirname(path) # only 1 level of sub folders, so this is safe
-          used << path
-          payload = part.as_json.merge(api_resource: part.class.api_resource)
-          write_file_if_necessary(path, JSON.pretty_generate(payload) << "\n")
+          Utils.parallel(parts, max: 2) do |part|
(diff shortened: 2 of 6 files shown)


Readme.md
```diff
@@ -248,6 +248,7 @@ so they can be created in a single update and can be re-created if any of them i
  - rebase on updated `master` to not undo other changes
  - figure out project name by converting the class name to snake-case
  - run `PROJECT=foo bundle exec rake kennel:update_datadog` to test changes for a single project (monitors: remove mentions while debugging to avoid alert spam)
+   - you can also use `PROJECT=foo:bar:...` if your work spans multiple projects
```
grosser, line 251: prefer to split by `,` since `:` already has a use in kennel

lib/kennel.rb
```diff
@@ -63,19 +63,21 @@ def update
...
-        used = []
-
-        Utils.parallel(parts, max: 2) do |part|
-          path = "generated/#{part.tracking_id.tr("/", ":").sub(":", "/")}.json"
-          used << File.dirname(path) # only 1 level of sub folders, so this is safe
-          used << path
-          payload = part.as_json.merge(api_resource: part.class.api_resource)
-          write_file_if_necessary(path, JSON.pretty_generate(payload) << "\n")
-        end
+        (project_filter || ["**"]).each do |project_pattern|
```
grosser, line 66: this would also work and not need another nesting ?
```
        old = Dir["generated/{#{project_filter || "**"}}/*"]
```

lib/kennel.rb
```diff
@@ -109,13 +111,13 @@ def generated
...
               known << kennel_id
-              next [] if kennel_id != project_filter
+              next [] unless project_filter.include?(kennel_id)
             end
             project.validated_parts
           end
 
           if project_filter && parts.empty?
-            raise "#{project_filter} does not match any projects, try any of these:\n#{known.uniq.sort.join("\n")}"
+            raise "#{project_filter_raw} does not match any projects, try any of these:\n#{known.uniq.sort.join("\n")}"
```
grosser, line 120: maybe better to use a single interface throughout:
```suggestion
            raise "#{project_filter.join(",")} does not match any projects, try any of these:\n#{known.uniq.sort.join("\n")}"
```

lib/kennel.rb
zdrve:
```ruby
irb(main):001:0> Dir["generated/#{['kube2iam', 'zeus_smoke']}/*"]
=> []
```

No, I don't think so.

Readme.md
zdrve: `,` feels more natural, whereas `:` with get the guarantee that the project id itself can't contain `:`.

Practically speaking if a project id doesn't match `/\A\w+\z/` then my answer would probably be "well don't do that then"

So sure, I'm fine with `,`

lib/kennel.rb
zdrve: It's not clear to me that either one is better than the other. Advantage of using `_raw`: encapsulation of the format - no need to ensure that this `join` is the reverse of the `split` elsewhere. Advantage of _not_ using `_raw`: smaller interface.

🤷‍♀️

grosser:
reworking this in a new pr 🤞

PR #155, opened 2022-04-15
Zdrve zdrve/multi project filter

grosser:
replaces https://github.com/grosser/kennel/pull/154


grosser:
also added some optimizations to not call project_filter method inside of loops

PR #158, opened 2022-04-27
Move lock to start

zdrve:
and add a space

Datadog object names are often shown in lists, and when they are, they're usually start-aligned (i.e. for English, left-aligned). I like it when things line up. Therefore, this PR moves the padlock emoji to the start of the text (and adds a space between the padlock and the rest of the object name), so that all the padlocks line up. Because that's pleasing, and less distracting.

### DIFF
--- a/lib/kennel/importer.rb
+++ b/lib/kennel/importer.rb
@@ -26,6 +26,7 @@ module Kennel
       title_field = TITLES.detect { |f| data[f] }
       title = data.fetch(title_field)
       title.tr!(Kennel::Models::Record::LOCK, "") # avoid double lock icon
+      title.strip!
 
       # calculate or reuse kennel_id
       data[:kennel_id] =
--- a/lib/kennel/models/dashboard.rb
+++ b/lib/kennel/models/dashboard.rb
@@ -161,7 +161,7 @@ module Kennel
 
         @json = {
           layout_type: layout_type,
-          title: "#{title}#{tags_as_string}#{LOCK}",
+          title: "#{LOCK} #{title}#{tags_as_string}",
           description: description,
           template_variables: render_template_variables,
           template_variable_presets: template_variable_presets,
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -58,7 +58,7 @@ module Kennel
       def as_json
         return @as_json if @as_json
         data = {
-          name: "#{name}#{LOCK}",
+          name: "#{LOCK} #{name}",
           type: type,
           query: query.strip,
           message: message.strip,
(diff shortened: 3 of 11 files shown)


zdrve:
If its not important then we should get rid of the lock entirely, and we'd both be happy.

zdrve:
>  Also the alignment between kennel and non-kennel is better when it's not in the front.

I don't understand this comment at all, I'm afraid.

grosser:
The lock is an indicator for "managed by kennel" we could use 🐶, but that's kinda ugly :D
... it's not a perfect indicator since sometimes it's wrong, removing it would make it harder to spot managed vs unmanaged though

> Also the alignment between kennel and non-kennel is better when it's not in the front.

If you have some things managed by kennel and others not they'd look like
```
Not managed
🔒 Managed
Not managed
🔒 Managed
```
which is kinda ugly too

zdrve:
> If you have some things managed by kennel and others not they'd look like...

The fix for that is to put the remaining things into kennel.

There is no equivalent fix if the padlock is at the end.

zdrve:
I would also suggest that it's _easier_ to see the locked-vs-not-locked, if the lock is always in the same place, i.e. column 1, the left.

grosser:
still feel like the end positions is nicer overall and less "in your face"
... not super married to it, but would be nice to get some broader feedback, maybe start a poll

tschroeder-zendesk:
Personally I like the end as well because we already have a hard time seeing what the monitors on our phones are due to length limitations on pagerduty. Adding the lock to the front means even more information we won't see in critical times.

mbigornia:
> The lock is an indicator for "managed by kennel"

Would it be a good idea to put a `kennel` or `managed_by:kennel` tag on the monitors instead of relying on the lock? This way we can remove the lock, and also search and filter more easily.

grosser:
The tag is a neat idea, it would not work for everything (like dashboards) but would make things easy to find. Not sure if finding things that are managed in kennel is common though, usually I want the opposite "show me things not in kennel".

zdrve:
`-tag:managed_by:kennel`, right?

grosser:
does that work ?

mbigornia:
Yep, that works - just tried :D

We can also extend that idea to other teams using different repos/monitor managers, so other monitors will have `managed_by:terraform`, etc. I feel this way we'll have a clearer view of how monitors are managed within Zendesk.

grosser:
Sounds good!

grosser:
hmmm what about dashboards though ?

grosser:
already complained about no tags on dashboards multiple times, but good to keep poking em ...
afaik they think "dashboard lists is good enough" 😞

grosser:
having a different approach for different resources is kinda meh too, hard to explain what is going on.
starting with adding the tag is definitely a good step an unlocks some new queries ...

PR #162, opened 2022-05-15
Parallelize project_class.new.validated_parts

zdrve:
Saves 1-2 seconds with the Zendesk data set on my machine.

Removed the "somehow" comment; it's not a mystery: it's CPU-intensive, and threading allows multiple CPUs to be used.

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -108,7 +108,7 @@ module Kennel
           known = []
           filter = project_filter
 
-          parts = Models::Project.recursive_subclasses.flat_map do |project_class|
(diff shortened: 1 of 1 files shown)


grosser:
I put the `somehow` there since afaik with the [GIL](https://jonathan-wong.medium.com/what-is-global-interpreter-lock-in-relation-to-ruby-18d0f4f20e20) I did not expect threads to be much faster, but I guess there is enough non-GIL work to be worth it.

grosser:
.. I kinda want to use parallel processes there, but this is still a good first step :)

grosser:
v1.113.1

PR #164, opened 2022-05-30
Add support for the 'silenced' option.

jamesbrauman:
The [Monitor API](https://docs.datadoghq.com/monitors/guide/monitor_api_options/) supports the 'silenced' option. This can be used to set the monitor be muted.

In practical usage when testing out new monitors I have built with `kennel` I want to be able to create the monitors in Datadog without creating spurious alerts.

I've also allowed to set the `silenced` option to `true` which isn't supported by the Monitor API. This is translated to `{ '*': nil }` which mutes the monitor completed.

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -23,7 +23,8 @@ module Kennel
         renotify_interval: 0,
         notify_audit: false,
         no_data_timeframe: nil, # this works out ok since if notify_no_data is on, it would never be nil
-        groupby_simple_monitor: false
+        groupby_simple_monitor: false,
+        silenced: nil,
       }.freeze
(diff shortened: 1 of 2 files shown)


jamesbrauman:
@grosser Please note I haven't tested this against the Monitor API as I wasn't sure how to, could you please either test or advise on how to test this. Thanks.

grosser (commented):
kennel does not support silencing in it's api to leave that to manual operation (for quick-fixes)
adding it to kennel would break that afaik since every monitor update would re-silence or un-silence monitor

to test a new monitor I'd recommend not setting a mention target (or setting the threshold high)

jamesbrauman:
That makes sense, thanks for the feedback.

PR #167, opened 2022-08-12
do not claim something comes from kennel

grosser:
still does not work right, gives things like `Rakefile` 😞 
... might have to go and fish for methods, but that's ugly too


grosser:
use const_source_location once >=2.7 is the default ruby

PR #168, opened 2022-08-12
Preserve source location when using non-procs

zdrve:
Discussion:

The "Managed by kennel" line adds the tracking id, and then the project class's `file_location`.  `file_location` in turn is worked out by picking an essentially random instance method from the class (`instance_methods(false).first`) and using a processed form of its source_location.

This was a problem for the new proc-less form because the source location of the procs was inside the Kennel codebase (see `AS_PROCS`). Furthermore, the results were unstable, because of the randomness of the chosen instance method; if it picked one which was defined _with_ a proc, everything was fine, but if it picked one using the new proc-less syntax, then the source_location and therefore file_location and therefore "Managed by kennel" line, would be wrong.

### DIFF
--- a/lib/kennel/models/project.rb
+++ b/lib/kennel/models/project.rb
@@ -11,6 +11,8 @@ module Kennel
       def self.file_location
         @file_location ||= begin
           method_in_file = instance_methods(false).first
+          return if method_in_file.nil?
+
           instance_method(method_in_file).source_location.first.sub("#{Bundler.root}/", "")
         end
       end
--- a/lib/kennel/settings_as_methods.rb
+++ b/lib/kennel/settings_as_methods.rb
@@ -4,11 +4,15 @@ module Kennel
     SETTING_OVERRIDABLE_METHODS = [].freeze
 
     AS_PROCS = ->(options) do
+      # Fragile; depends on the fact that in both locations where AS_PROCS is
+      # used, we're 2 frames away from the user code.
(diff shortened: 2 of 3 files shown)


grosser (approved):
🎉 much simpler :D

grosser:
rubocop would like a work with you :)
... a test would also be nice so we see when we break it

lib/kennel/models/project.rb
```diff
@@ -11,6 +11,8 @@ class Project < Base
       def self.file_location
         @file_location ||= begin
           method_in_file = instance_methods(false).first
+          return if method_in_file.nil?
```
grosser, line 14: that should never happen now right ?

grosser:
1.114.1

grosser:
... we should also check this new syntax performance wise just in case it's super slow

lib/kennel/models/project.rb
zdrve: Super edge case, will never happen with normal data. See the test case.

zdrve:
> in case it's super slow

Seems unlikely; the old syntax made procs and passed them to kennel; the new syntax passes values to kennel, which then makes procs. But yes it is of course an extra opcode or two just to do the detection.

grosser:
FYI ran a small benchmark and there is no noticeable difference (loading 2k monitors with old or new syntax)

PR #173, opened 2022-08-26
Use the call stack to Record#new in the "Managed by kennel" line

zdrve:
near-duplicate of https://github.com/zdrve/kennel/pull/1 

(this one runs the tests; the other one has the right base branch)

### DIFF
--- a/lib/kennel/models/project.rb
+++ b/lib/kennel/models/project.rb
@@ -17,13 +17,13 @@ module Kennel
         end
       end
 
-      def validated_parts
+      def validated_parts(base_dir)
         all = parts
         unless all.is_a?(Array) && all.all? { |part| part.is_a?(Record) }
           invalid! "#parts must return an array of Records"
         end
 
-        all.each { |part| part.strip_caller(here) }
+        all.each { |part| part.strip_caller(base_dir) }
         validate_parts(all)
         all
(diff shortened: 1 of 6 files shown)


lib/kennel/models/project.rb
```diff
@@ -18,7 +18,10 @@ def self.file_location
       end
 
       def validated_parts
-        all = parts
+        # 'parts' called on same line as 'caller' for maximum cleanliness in strip_caller
+        here, all = [->() {caller}.call, parts]
+        all = [all] unless all.is_a?(Array)
```
grosser, line 23: can do this or `all ||= []`
```suggestion
        all = Array(all)
```

grosser, line 23:
```suggestion
        here, all = [->() {caller}.call, parts.to_a]
```

grosser: might be simpler to say `remove everything before __FILE__:__LINE__`

zdrve: The `all = [all]` was there for a purpose... however, this https://github.com/zendesk/kennel/pull/14227 (plus a follow-up grosser-kennel PR to fail if `parts` returns a non-array) is the better fix for that problem.

PR #174, opened 2022-08-27
`#parts` must return an array

zdrve:
See also https://github.com/zendesk/kennel/pull/14227 

This is a breaking change (insofar as it breaks projects where `parts` returns a record, not an array)

### DIFF
--- a/lib/kennel/models/project.rb
+++ b/lib/kennel/models/project.rb
@@ -2,6 +2,12 @@
 module Kennel
   module Models
     class Project < Base
+      class PartsDidNotReturnAnArrayException < RuntimeError
+        def initialize(project)
+          super("Project#parts (in #{project.name} / #{project.kennel_id}) must return an array")
+        end
+      end
+
       settings :team, :parts, :tags, :mention, :name, :kennel_id
       defaults(
         tags: -> { ["service:#{kennel_id}"] + team.tags },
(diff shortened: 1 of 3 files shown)


lib/kennel/models/project.rb
```diff
@@ -19,6 +25,8 @@ def self.file_location
 
       def validated_parts
         all = parts
+        raise PartsDidNotReturnAnArrayException, self unless all.is_a?(Array)
```
grosser, line 28:
```suggestion
        invalid! "parts must return an array" unless all.is_a?(Array)
```

test/kennel/models/project_test.rb
```diff
@@ -66,5 +66,14 @@ def self.to_s
     it "returns parts" do
       TestProject.new.validated_parts.size.must_equal 0
     end
+
+    it "raises an error if parts did not return an array" do
+      bad_project = TestProject.new(parts: -> {
+        Kennel::Models::Monitor.new(self)
+      })
+      validation_error_message do
```
grosser, line 74: uh nice!

lib/kennel/models/project.rb
```diff
@@ -19,12 +19,21 @@ def self.file_location
...
+
         validate_parts(all)
         all
       end
 
       private
 
+      # let users know which project/resource failed when something happens during diffing where the backtrace is hidden
+      def invalid!(message)
+        raise ValidationError, "#{kennel_id} #{message}"
```
grosser, line 34: ah I thought that existed :D

grosser (approved):
nice!

lib/kennel/models/project.rb
zdrve: Only for Record, not for Project.

PR #175, opened 2022-08-28
Thread safety

zdrve:
_progress_ is now thread-safe.

_parallel_ is _closer to_ thread-safety, but it's still not safe to assume that it's thread-safe.

It's going to take quite a lot of work to make it _actually_ thread-safe.

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -67,11 +67,14 @@ module Kennel
       Progress.progress "Storing" do
         old = Dir["generated/{#{(project_filter || ["**"]).join(",")}}/*"]
         used = []
+        mutex = Mutex.new
 
         Utils.parallel(parts, max: 2) do |part|
           path = "generated/#{part.tracking_id.tr("/", ":").sub(":", "/")}.json"
-          used << File.dirname(path) # only 1 level of sub folders, so this is safe
-          used << path
+          mutex.synchronize do
+            used << File.dirname(path) # only 1 level of sub folders, so this is safe
+            used << path
+          end
           payload = part.as_json.merge(api_resource: part.class.api_resource)
           write_file_if_necessary(path, JSON.pretty_generate(payload) << "\n")
         end
@@ -107,12 +110,13 @@ module Kennel
           load_all
           known = []
           filter = project_filter
+          mutex = Mutex.new
 
           parts = Utils.parallel(Models::Project.recursive_subclasses) do |project_class|
             project = project_class.new
             kennel_id = project.kennel_id
             if filter
-              known << kennel_id
+              mutex.synchronize { known << kennel_id }
               next [] unless filter.include?(kennel_id)
             end
             project.validated_parts
--- a/lib/kennel/progress.rb
+++ b/lib/kennel/progress.rb
@@ -7,24 +7,25 @@ module Kennel
     def self.progress(name)
       Kennel.err.print "#{name} ... "
 
-      animation = "-\\|/"
-      count = 0
       stop = false
       result = nil
+      mutex = Mutex.new
 
       spinner = Thread.new do
+        animation = "-\\|/"
+        count = 0
         loop do
-          break if stop
-          Kennel.err.print animation[count % animation.size]
+          break if mutex.synchronize { stop }
+          Kennel.err.print animation[count]
           sleep 0.2
           Kennel.err.print "\b"
-          count += 1
+          count = (count + 1) % animation.size
         end
       end
 
       time = Benchmark.realtime { result = yield }
 
-      stop = true
+      mutex.synchronize { stop = true }
       spinner.join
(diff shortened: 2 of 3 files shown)


lib/kennel.rb
```diff
@@ -107,12 +110,13 @@ def generated
...
           known = []
           filter = project_filter
+          mutex = Mutex.new
 
           parts = Utils.parallel(Models::Project.recursive_subclasses) do |project_class|
             project = project_class.new
             kennel_id = project.kennel_id
             if filter
-              known << kennel_id
+              mutex.synchronize { known << kennel_id }
```
grosser, line 119: array access is thread-safe in ruby, so we don't need this

lib/kennel.rb
```diff
@@ -67,11 +67,14 @@ def store(parts)
...
+        mutex = Mutex.new
 
         Utils.parallel(parts, max: 2) do |part|
           path = "generated/#{part.tracking_id.tr("/", ":").sub(":", "/")}.json"
-          used << File.dirname(path) # only 1 level of sub folders, so this is safe
-          used << path
+          mutex.synchronize do
+            used << File.dirname(path) # only 1 level of sub folders, so this is safe
+            used << path
+          end
```
grosser, line 77: prefer concat which is thread-safe
```suggestion
          used.concat [File.dirname(path), path] # only 1 level of sub folders, so this is safe (use concat for thread-safety)
```

lib/kennel/progress.rb
```diff
@@ -7,24 +7,25 @@ class Progress
...
       result = nil
+      mutex = Mutex.new
 
       spinner = Thread.new do
+        animation = "-\\|/"
+        count = 0
         loop do
-          break if stop
-          Kennel.err.print animation[count % animation.size]
+          break if mutex.synchronize { stop }
```
grosser, line 18: we never un-set it, so no need to use a mutex ?

lib/kennel/utils.rb
```diff
@@ -105,18 +105,22 @@ def parallel(items, max: 10)
         threads = [items.size, max].min
         work = items.each_with_index.to_a
         done = Array.new(items.size)
+        mutex = Mutex.new
         workers = Array.new(threads).map do
           Thread.new do
             loop do
-              item, i = work.pop
+              item, i = mutex.synchronize { work.pop }
```
grosser, line 112: pop is thread-safe

lib/kennel/progress.rb
```diff
@@ -7,24 +7,25 @@ class Progress
...
 
-      animation = "-\\|/"
-      count = 0
       stop = false
       result = nil
+      mutex = Mutex.new
 
       spinner = Thread.new do
+        animation = "-\\|/"
+        count = 0
```
grosser, line 16: keeping as much context as possible inside the thread is good 👍

lib/kennel/utils.rb
```diff
@@ -105,18 +105,22 @@ def parallel(items, max: 10)
...
                 begin
                   yield item
                 rescue StandardError => e
-                  work.clear
                   e
                 end
+              mutex.synchronize do
+                done[i] = result
+                work.clear if e.is_a?(StandardError)
+              end
```
grosser, line 123: it does not matter when we clear / there is no race condition afaik

grosser:
merged the generate fixes into master https://github.com/grosser/kennel/commit/6f1bd67646032aa4fde47998307f665630ce6354

lib/kennel/utils.rb
zdrve: Citation? 

I've just dug around a bit in the ruby documentation and failed to find anything backing up your assertions about things being thread safe.

lib/kennel/utils.rb
zdrve: And yet kennel gets stuck when running the github actions – only the ones involving this threading.

grosser:
```
workers.each(&:join)
lib/kennel/utils.rb:123:in `join'
```
from 
```
parts = Utils.parallel(projects, &:validated_parts).flatten(1)
lib/kennel.rb:121
```

... and in the thread it's

```
gems/zeitwerk-2.6.0/lib/zeitwerk/kernel.rb:27:in `require'
```

so something about auto-load
the other sus thing was that we rescue StandardError and not Exception but did not set abort_on_exception so good chance there is something going on there too
fix for that https://github.com/grosser/kennel/pull/177

still it seems like the slow tests hang in random places, so maybe there is something eating all the cpu and that is why it's slow

grosser:
confirmed that the issue is with zeitwerk ...
switching to eager-load produces no more errors whereas before I was able to consistently reproduce

grosser:
fix https://github.com/grosser/kennel/pull/178

PR #179, opened 2022-09-01
Implement a method to print diffs using git-diff.

jamesbrauman:


lib/kennel/plans/printers/git.rb
```diff
@@ -0,0 +1,86 @@
...
+          contents = Utils.sorted_hash_enumeration_order(contents)
+
+          JSON.pretty_generate(contents)
+        end
+
+        def execute_diff!(parent_directory, from_directory, to_directory)
+          from_relative = from_directory.relative_path_from(parent_directory)
+          to_relative = to_directory.relative_path_from(parent_directory)
+
+          command = DIFF_COMMAND % [parent_directory, from_relative, to_relative]
```
zdrve, line 64: Shell quoting vulnerabilities. Use system with multiple args and the cwd (iirc) option instead.

lib/kennel/plans/printers/git.rb
```diff
@@ -0,0 +1,86 @@
...
+
+          directories.each(&Dir.method(:mkdir))
+
+          [parent_directory] + directories
+        end
+
+        def delete_temporary_directories(*directories)
+          directories.each { |directory| FileUtils.remove_entry(directory) }
+        rescue Errno::ENOENT
+          # Ignored
```
zdrve, line 81: This can leave directories behind

lib/kennel/plans/printers/git.rb
```diff
@@ -0,0 +1,86 @@
...
+
+require "open3"
+require "shellwords"
+require "tmpdir"
+
+module Kennel
+  module Plans
+    module Printers
+      class Git
+        DIFF_COMMAND = "(cd %s && git --no-pager diff --no-prefix --color=always --no-index %s %s)"
```
zdrve, line 11: Why git diff, not just diff?

lib/kennel/plans/printers/git.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        attr_reader :plan
+
+        def initialize(plan, io)
+          @plan = plan
+          @io = io
+        end
+
+        def print!
+          parent_directory, from_directory, to_directory = create_temporary_directories
```
zdrve, line 21: The whole business with temporary directories and file-writing can potentially be made much simpler by using temporary files or pipes.

zdrve:
Aside from the implementation issues, I'm not quite sure what problem this is trying to solve.

Could you perhaps provide a screenshot of it in action, maybe also with some text explaining why it's useful?

jamesbrauman:
@zdrve  - Thanks for the feedback - much appreciated! After looking into `git-diff` more, I found that it was too intricately tied to `git` for my use case here.

Instead I've opened https://github.com/grosser/kennel/pull/180 - keen to hear your thoughts!

PR #181, opened 2022-09-02
Atomic write in FileCache

zdrve:
I hit a problem with a corrupted cache earlier; I didn't dig too deeply, but my guess is that it was caused by non-atomic writes.

### DIFF
--- a/lib/kennel/file_cache.rb
+++ b/lib/kennel/file_cache.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'tempfile'
+
 # cache that reads everything from a single file
 # - avoids doing multiple disk reads while iterating all definitions
 # - has a global expiry to not keep deleted resources forever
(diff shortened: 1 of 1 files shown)


grosser (commented):
this is technically correct, but sadly slow and does lots of disk IO to fix a 0.x% case
```
require 'tempfile'
require 'benchmark'

c = Dir["generated/**/*"].size
puts "MOVE"
puts(Benchmark.realtime do
  c.times do
    Tempfile.open('FileCache', "/tmp") do |tmp|
      tmp.write "test" * 100
      File.chmod 0o600, tmp.path
      File.rename tmp.path, "foo"
    end
  end
end)

puts "PLAIN"
puts(Benchmark.realtime do
  c.times do
    File.write("foo", "test" * 100)
  end
end)
-->
13084
MOVE
2.7888690000399947
PLAIN
0.9394689999753609
```

so prefer doing a rescue on read or some maybe use a rescue to delete the file if anything blows up

grosser (approved):
ahh this is not the generated path, so this is fine to be slow :D

grosser:
manually merging

zdrve:
Also, regardless of speed: it's not important on the generated path, because generated files are not read (apart from the only-write-if-different logic). So corruption does not persist.

PR #183, opened 2022-09-03
use diff-lcs for long text

grosser:
SCR-20220902-r46


zdrve:
How does this relate to #180?

zdrve:
Ah no worries, just seen the comment over there. 👍

zdrve (approved):

PR #160, opened 2022-05-06
Allow id-expansion in description/message

zdrve:
Allow `%{TYPE:kennel_id_with_colon}` (e.g. `%{dashboard:myproject:error_budgets}`) in description/message, to expand to the datadog-id of that item.

This allows us to write code like this, in e.g. a SLO / monitor message:

```ruby
dashboard = Sre::Dashboards::ErrorBudgetExplorer.new(Sre::Project.new)
dashboard_url = "https://zendesk.datadoghq.com/dashboard/#{dashboard.id_macro}/error-budget-explorer?tpl_var_metric=#{sli.name}&tpl_var_source=#{sli.capture_source}"

"If this alarm fires, go and look at [the dashboard](#{dashboard_url})" ...
```

No tests yet.

PR #184, opened 2022-09-03
Smarter update diff simpler

zdrve:
Have `.plan` and `.update` return useful data structures.

Unrelated, though useful for testing: allow the `Managed by kennel` marker to be overridden.

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -42,8 +42,8 @@ module Kennel
 
   include Kennel::Compatibility
 
-  PlanResults = Struct.new(:plan, keyword_init: true)
-  UpdateResults = Struct.new(:plan, :update, keyword_init: true)
+  PlanResult = Struct.new(:plan, keyword_init: true)
+  UpdateResult = Struct.new(:plan, :update, keyword_init: true)
 
   class Engine
     def initialize
@@ -62,7 +62,7 @@ module Kennel
 
     def plan
       the_plan = syncer.plan
-      PlanResults.new(
+      PlanResult.new(
         plan: the_plan
       )
     end
@@ -70,7 +70,7 @@ module Kennel
     def update
       the_plan = syncer.plan
       the_update = syncer.update if syncer.confirm
-      UpdateResults.new(
+      UpdateResult.new(
         plan: the_plan,
         update: the_update
       )
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -54,11 +54,6 @@ module Kennel
 
     attr_accessor :out, :err, :strict_imports
 
-    def reset
-      remove_instance_variable(:@generated)
-      remove_instance_variable(:@syncer)
-    end
-
     def generate
       out = generated
       store out if ENV["STORE"] != "false" # quicker when debugging
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -175,13 +175,21 @@ module Kennel
     end
 
     def load_all
-      @loaded_zeitwerk ||= begin
-        loader = Zeitwerk::Loader.new
-        Dir.exist?("teams") && loader.push_dir("teams", namespace: Teams)
-        Dir.exist?("parts") && loader.push_dir("parts")
-        loader.setup
-        loader.eager_load # TODO: this should not be needed but we see hanging CI processes when it's not added
-      end
+      # load_all's purpose is to "require" all the .rb files under './projects',
+      # also with reference to ./teams and ./parts. What happens if you call it
+      # more than once?
+      #
+      # For a reason yet to be investigated, Zeitwerk rejects second and subsequent calls.
+      # But even if we skip over the Zeitwerk part, the nature of 'require' is
+      # somewhat one-way: we're not providing any mechanism to *un*load things.
+      # As long as the contents of `./projects`, `./teams` and `./parts` doesn't
+      # change between calls, then simply by no-op'ing subsequent calls to `load_all`
+      # we can have `load_all` appear to be idempotent.
+      loader = Zeitwerk::Loader.new
+      Dir.exist?("teams") && loader.push_dir("teams", namespace: Teams)
+      Dir.exist?("parts") && loader.push_dir("parts")
+      loader.setup
+      loader.eager_load # TODO: this should not be needed but we see hanging CI processes when it's not added
 
       # TODO: also auto-load projects and update expected path too
       ["projects"].each do |folder|
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -42,7 +42,6 @@ module Kennel
 
   include Kennel::Compatibility
 
-  PlanResult = Struct.new(:plan, keyword_init: true)
   UpdateResult = Struct.new(:plan, :update, keyword_init: true)
 
   class Engine
@@ -61,10 +60,7 @@ module Kennel
     end
 
     def plan
-      the_plan = syncer.plan
-      PlanResult.new(
-        plan: the_plan
-      )
+      syncer.plan
     end
 
     def update
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -42,6 +42,9 @@ module Kennel
 
   include Kennel::Compatibility
 
+  PlanResults = Struct.new(:plan, keyword_init: true)
+  UpdateResults = Struct.new(:plan, :update, keyword_init: true)
+
   class Engine
     def initialize
       @out = $stdout
@@ -51,6 +54,11 @@ module Kennel
 
     attr_accessor :out, :err, :strict_imports
 
+    def reset
+      remove_instance_variable(:@generated)
+      remove_instance_variable(:@syncer)
+    end
+
     def generate
       out = generated
       store out if ENV["STORE"] != "false" # quicker when debugging
@@ -58,12 +66,19 @@ module Kennel
     end
 
     def plan
-      syncer.plan
+      the_plan = syncer.plan
+      PlanResults.new(
+        plan: the_plan
+      )
     end
 
     def update
-      syncer.plan
-      syncer.update if syncer.confirm
+      the_plan = syncer.plan
+      the_update = syncer.update if syncer.confirm
+      UpdateResults.new(
+        plan: the_plan,
+        update: the_update
+      )
     end
 
     private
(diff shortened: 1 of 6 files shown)


lib/kennel/tasks.rb
```diff
@@ -71,18 +109,7 @@ def abort(message = nil)
...
-    task_name =
-      if on_default_branch && is_push
-        Kennel.strict_imports = false
-        "kennel:update_datadog"
-      else
-        "kennel:plan" # show plan in CI logs
-      end
-
-    Rake::Task[task_name].invoke
+    Kennel::Tasks.ci
```
zdrve, line 112: Come to think of it, I'm not sure that `:ci` needed to move out of rake and into a normal method. But it does no harm.

lib/kennel/syncer.rb
```diff
@@ -55,8 +70,11 @@ def update
         message = "#{klass.api_resource} #{a.fetch(:tracking_id)} #{id}"
         Kennel.out.puts "Deleting #{message}"
         @api.delete klass.api_resource, id
+        update_log << [:delete, klass.api_resource, id]
         Kennel.out.puts "#{LINE_UP}Deleted #{message}"
       end
+
+      Update.new(update_log: update_log)
```
zdrve, line 77: Note that `update_log` exposes quite a bit of data (and all undocumented). For the purposes of the corresponding zendesk kennel PR we _could_ have made it expose just the tracking ids.

lib/kennel.rb
```diff
@@ -58,12 +61,19 @@ def generate
     end
 
     def plan
-      syncer.plan
+      the_plan = syncer.plan
+      PlanResults.new(
+        plan: the_plan
+      )
```
grosser, line 67:
```suggestion
      PlanResults.new(plan: syncer.plan)
```
...  but all this does is add indirection, so prefer keeping:
```
syncer.plan
```

lib/kennel.rb
```diff
@@ -58,12 +61,19 @@ def generate
...
     end
 
     def update
```
grosser, line 70: could `update` maybe get a plan ?
```suggestion
    def update(plan)
```

grosser: hmmm I think this will need more refactor since there is lots of instance state ...

lib/kennel.rb
```diff
@@ -58,12 +61,19 @@ def generate
...
-      syncer.plan
-      syncer.update if syncer.confirm
+      the_plan = syncer.plan
+      the_update = syncer.update if syncer.confirm
+      UpdateResults.new(
```
grosser, line 73: singular since it is 1 result ?
```suggestion
      UpdateResult.new(
```

lib/kennel.rb
```diff
@@ -169,11 +179,13 @@ def filter_resources!(resources, by, against, name, env)
     end
 
     def load_all
-      loader = Zeitwerk::Loader.new
-      Dir.exist?("teams") && loader.push_dir("teams", namespace: Teams)
-      Dir.exist?("parts") && loader.push_dir("parts")
-      loader.setup
-      loader.eager_load # TODO: this should not be needed but we see hanging CI processes when it's not added
+      @loaded_zeitwerk ||= begin
```
grosser, line 182: add a comment explaining why we need to cache zeitwerk but re-run the require

lib/kennel/models/record.rb
```diff
@@ -2,6 +2,7 @@
 module Kennel
   module Models
     class Record < Base
+      MANAGED_BY_KENNEL = ENV.fetch("MANAGED_BY_KENNEL", "Managed by kennel")
```
grosser, line 5:
```suggestion
      MANAGED_MARKER = ENV.fetch("KENNEL_MANAGED_MARKER", "Managed by kennel")
```

lib/kennel/models/record.rb
```diff
@@ -92,11 +93,11 @@ def resolve_linked_tracking_ids!(*)
       def add_tracking_id
         json = as_json
         if self.class.parse_tracking_id(json)
-          invalid! "remove \"-- Managed by kennel\" line it from #{self.class::TRACKING_FIELD} to copy a resource"
+          invalid! "remove \"-- #{MANAGED_BY_KENNEL}\" line it from #{self.class::TRACKING_FIELD} to copy a resource"
         end
         json[self.class::TRACKING_FIELD] =
           "#{json[self.class::TRACKING_FIELD]}\n" \
-          "-- Managed by kennel #{tracking_id} in #{project.class.file_location}, do not modify manually".lstrip
+          "-- #{MANAGED_BY_KENNEL} #{tracking_id} in #{project.class.file_location}, do not modify manually".lstrip
```
grosser, line 100: good refactor overall ... not sure why we need the env setter, so comment at the const definition would be nice to explain why it's useful

lib/kennel/syncer.rb
```diff
@@ -27,6 +30,14 @@ def plan
         print_plan "Update", @update, :yellow
         print_plan "Delete", @delete, :red
       end
+
+      Plan.new(
+        noop?: noop?,
+        already_ok: @already_ok,
```
grosser, line 36: these seem a bit redundant, maybe a plan class could hold the attrs and have a `noop?` method ?

lib/kennel/syncer.rb
grosser: how is the update-log different from a `Plan` ? ... it has the ids resolved ?
can we use the `Plan` internally and then resolve the ids in the plan to have the same data structure at the end ?

lib/kennel/syncer.rb
```diff
@@ -133,7 +152,11 @@ def calculate_diff
           id = a.fetch(:id)
           if e
             diff = e.diff(a) # slow ...
-            @update << [id, e, a, diff] if diff.any?
+            if diff.any?
+              @update << [id, e, a, diff]
+            else
+              @already_ok << [id, e, a]
```
grosser, line 158: better name ?
```suggestion
              @unchanged << [id, e, a]
```

lib/kennel/tasks.rb
```diff
@@ -12,6 +12,44 @@ def abort(message = nil)
         Kennel.err.puts message if message
         raise SystemExit.new(1), message
       end
+
+      def environment
+        @environment ||= begin
+          require "kennel"
```
grosser, line 18: maybe cleaner ?
```suggestion
        return if @environment_loaded
        @environment_loaded = true
        require "kennel"
```

lib/kennel/tasks.rb
```diff
@@ -12,6 +12,44 @@ def abort(message = nil)
         Kennel.err.puts message if message
         raise SystemExit.new(1), message
       end
+
+      def environment
```
grosser, line 16: ... also `def load_environment` or `def setup_environment`

lib/kennel/tasks.rb
grosser: would rather keep it then, to reduce indirection, but either way is fine

lib/kennel.rb
zdrve: Because it doesn't work otherwise?

I know that's not a very helpful answer. I don't really understand, or trust, Zeitwerk.

lib/kennel/tasks.rb
zdrve: I'd rather keep it where it is – class/module code is easier to reuse / refactor / test. Rake task code should ideally be thin layer over the top of that.

lib/kennel.rb
zdrve: Small changes, future PRs. Yes the instance state in kennel is pretty messy to work with. I have a vague plan for how to address it, and in no way does it fit in this PR.

lib/kennel.rb
grosser: Something like "we load this multiple times when doing X"
also putting the require into the block, or doing a
```
return if @loaded
@loaded = true
```
would not re-run the require and make it a bit cleaner ?

zdrve: IIRC the error was something to do with there already being a loader registered. 🤷‍♀️

lib/kennel/syncer.rb
zdrve: Well.... at the moment, perhaps not much. My longer term view however is this:

The way that `plan` works at the moment – and therefore what the plan actually _is_ – is very simplistic. I prefer to think of it not so much of a plan, but more of a diff; a comparison. If we compare A to B, then this is added, this is changed, this is removed.

If the job then is resolve those differences (assuming it's a non-empty set), then we need to _do_ something. What, exactly? Well, that's the plan. Or, once performed, the results.

Our methodology is already showing its shortcomings. If we for example create a monitor M with tracking id K and type T, and then we update kennel so that the type now needs to be T', then (IIRC) Kennel currently can't handle that, because Datadog does not allow a monitor's type to be changed. However, it is solvable: instead of updating the monitor, delete monitor M then create monitor M' type T' (and with the same tracking ID K). In this case, the _diff_ would be "Monitor K change T => T'", but the _plan_ would be "delete M and create M'".

Another example: circular references. The reason that I haven't yet merged https://github.com/grosser/kennel/pull/160 is that it would allow for / increase the likelihood of circular references. Monitor M, whose message text includes a link to troubleshooting dashboard D. But dashboard D contains a view of monitor M. So if M => D and D => M, and neither exist, so the diff is "both D and M are new", what's the plan? Possible answer: create M (but with the message text _only_ containing the kennel tracking text); create D, referring to M; update M to set its message text correctly, linking to D.

To return to your question, "what's the difference between the plan and the update log?". What we currently call `plan` is a diff. What I'm calling update_log is closer to a plan – but retrospective.

lib/kennel.rb
zdrve: Sure. This name is internal to the kennel engine anyway.

lib/kennel/syncer.rb
zdrve: eehhhhhh not quite sure. See my long comment elsewhere about how I think of plan vs diff.

lib/kennel/tasks.rb
zdrve: Less safe IMO

lib/kennel.rb
grosser: yeah, so the method gets called twice, we should find out why and either stop it or make the whole method re-executable (the require still costs time and is wasteful to execute twice)

zdrve: Sure. That's not something that gets _worse_ with this PR though, right? This PR simply makes the problem more _visible_.

grosser: I'd say this PR makes the problem less visible since it allows ignoring it.
Before the invalid usage would fail.
I'd prefer to either avoid the invalid usage or make it not invalid anymore.
(as in: don't call this code twice or make the whole code hide behind `return if @loaded`, but not making half the code not execute)

lib/kennel/syncer.rb
zdrve: Maybe. In a later refactor, I would suggest.

zdrve: You're not suggesting changing the interface, just the implementation.

lib/kennel/syncer.rb
zdrve: Went with "no_change"

grosser:
need a release or more coming ?

zdrve:
In terms of the "smarter diff" feature, that's all that's needed. Smaller releases are better than big anyway :-)

grosser:
1.118.0

PR #185, opened 2022-09-05
Instance

ghost:
Move away from `Kennel.plan` etc to `Kennel::Engine.new.plan`

Although not included in this PR, it paves the way for:

 * passing the various options to the constructor, and making them immutable, instead of action-at-a-distance via `ENV`
 * easier resetting between tests
 * documenting the options
 * removes the need for "reset" in https://github.com/grosser/kennel/pull/184/files

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -63,7 +63,7 @@ Minitest::Test.class_eval do
     let(:stdout) { StringIO.new }
     let(:stderr) { StringIO.new }
 
-    before do |_t|
+    before do
       Kennel.out = stdout
       Kennel.err = stderr
     end


test/test_helper.rb
```diff
@@ -55,14 +63,12 @@ def self.capture_all
     let(:stdout) { StringIO.new }
     let(:stderr) { StringIO.new }
 
-    around do |t|
+    before do |_t|
```
grosser, line 66: FYI don't need the `_t`
```suggestion
    before do
```

grosser (commented):
looks alright, in itself it seems to just add complexity but I get where you are trying to take this

I'm hoping there is a better word than `Engine`, maybe extracting generator would help.
Also Engine could be just `config` and then hold no functionality, just be config parsing + storage.

grosser:
can you make a super-rough/untested/not-working PR with what the end-state is supposed to look like, so it's more obvious if this refactor is going in the right direction ?

zdrve:
Not sure how much of an "impression" vs "pseudocode" you're after: https://github.com/zdrve/kennel/pull/2/files

Also some future-direction thoughts which don't _quite_ fit into this sketch, except that I believe that this sort of work (Instance etc) would make the future directions easier to follow.

grosser:
yeah seems interesting ... this is good to merge then ?

grosser:
need a release or got more coming ?

PR #186, opened 2022-09-06
build options

zdrve:


grosser:
are we planing to have more than one of these ?
... because if not then we could just do `class Generator` or so

zdrve:
Not sure yet. Even splitting it out gives benefits, for testing etc; adding more than one implementation _might_ be overkill. Draft! ;-)

PR #187, opened 2022-09-15
s/environment/load_environment/

zdrve:
Whoops.

Stuff in `Tasks` has very poor test coverage, so this got missed.

The `ci_and_notify` implementation that Zendesk kennel has at the moment invokes `kennel:ci` and therefore hits this bug. That forthcoming `ci_and_notify` does _not_ use the `kennel:ci` task, and so actually this bug would have been avoided – or another way of saying that is, not noticed.

### DIFF
--- a/lib/kennel/tasks.rb
+++ b/lib/kennel/tasks.rb
@@ -32,7 +32,7 @@ module Kennel
       end
 
       def ci
-        environment
+        load_environment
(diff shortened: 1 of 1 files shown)


grosser (approved):

grosser:
1.118.1

PR #188, opened 2022-09-16
Zdrve/config

zdrve:


lib/kennel/compatibility.rb
```diff
@@ -4,8 +4,23 @@ module Kennel
   module Compatibility
     def self.included(into)
       class << into
-        %I[out out= err err= strict_imports strict_imports= generate plan update].each do |sym|
-          define_method(sym) { |*args| instance.public_send(sym, *args) }
+        %I[out err strict_imports].each do |sym|
+          define_method(sym) do |*args|
+            warn "Using legacy Kennel.#{sym} from #{caller[1..1]}"
```
grosser, line 9:
```suggestion
            warn "Using legacy Kennel.#{sym} from #{caller[1..1]}, use X instead"
```

lib/kennel.rb
```diff
@@ -45,13 +46,11 @@ class ValidationError < RuntimeError
   UpdateResult = Struct.new(:plan, :update, keyword_init: true)
 
   class Engine
-    def initialize
-      @out = $stdout
-      @err = $stderr
-      @strict_imports = true
+    def initialize(config_from = Kennel::Config.from_env, &block)
+      @config = Kennel::Config.new(config_from, &block)
```
grosser, line 50: passing a config instance to config.new seems strange, can we pass in a config instead ?
```suggestion
    def initialize(config = Kennel::Config.from_env)
      @config = config
```

lib/kennel/config.rb
```diff
@@ -0,0 +1,110 @@
...
+#   config = Kennel::Config.new(from)
+#
+# A block can be passed to `.new` or `.from_env`. If a block is given,
+# then after any settings have been copied from `from` (if given),
+# a mutable config object will be yielded to the block, so that
+# the config can be set up:
+#
+#   config = Kennel::Config.new do |c|
+#     c.strict_imports = false
+#   end
```
grosser, line 31: if someone wants to modify the config they can create their own object and pass it in ?

lib/kennel/config.rb
```diff
@@ -0,0 +1,110 @@
...
+#
+# Once built, a config object is immutable.
```
grosser, line 33: not a fan of that, makes testing/experimenting harder 🤷

lib/kennel/config.rb
```diff
@@ -0,0 +1,110 @@
...
+module Kennel
+  class Config
+    ATTRS = %I[
+      out
+      err
+      strict_imports
+    ].freeze
+    private_constant :ATTRS
+
+    attr_reader(*ATTRS)
```
grosser, line 50: just use an `attr_accessor` ?

lib/kennel.rb
zdrve: You can pass in anything which responds to `to_h`.

lib/kennel/config.rb
zdrve: Yup.

lib/kennel/config.rb
zdrve: I have the exact opposite viewpoint. Mutability makes testing harder. Don't like the config? Make a new one.

lib/kennel/config.rb
zdrve: No thanks. No validation.

lib/kennel/config.rb
grosser: if it's a global object then yeah, if it's per-instance then I don't really see the benefit since there would be a new one per test anyways
(no big deal either way)

lib/kennel/config.rb
grosser: I meant: we can remove the block interface and do `Config.new strict_imports: false` to have less api surface makes the input easier to test since we can have a "base" hash that gets modified each test instead of having to deal with a block

lib/kennel.rb
grosser: if someone needs to configure the config they can do it outside, and we can remove the `block` interface for initialize
can even make config a keyword-arg to leave room for future additions

zdrve: Yielding a mutable config object during build is a very common style that I'd like to keep.

grosser: The most common for I see is keyword-args (also easy to build/debug etc), if things get more complicated then yielding+config object might make sense though.
Kennel is for users that don't necessarily know a lot of ruby, so keeping things simple ("pass in a hash" vs "yield this config object") is nice.

lib/kennel/compatibility.rb
```diff
@@ -4,8 +4,11 @@ module Kennel
   module Compatibility
     def self.included(into)
       class << into
-        %I[out out= err err= strict_imports strict_imports= generate plan update].each do |sym|
-          define_method(sym) { |*args| instance.public_send(sym, *args) }
+        %I[generate plan update].each do |sym|
```
grosser, line 7: FYI actually prefer the simpler:
```suggestion
        [:generate, :plan, :update].each do |sym|
```

lib/kennel/tasks.rb
```diff
@@ -35,10 +35,12 @@ def ci
         load_environment
 
         if on_default_branch? && git_push?
-          Kennel.strict_imports = false
-          Kennel.update
+          config = Kennel::Config.from_env do |c|
+            c.strict_imports = false
+          end
```
grosser, line 40: another neat api could be:
```suggestion
          config = Kennel::Config.from_env.merge(strict_imports: false)
```

PR #190, opened 2022-09-24
Refactor out 'filter'

zdrve:
Also one bug fix. See its own commit.

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -79,10 +79,10 @@ module Kennel
       Progress.progress "Storing" do
         old = Dir[[
           "generated",
-          if project_filter || tracking_id_filter
+          if filter.project_filter || filter.tracking_id_filter
             [
-              "{" + (project_filter || ["*"]).join(",") + "}",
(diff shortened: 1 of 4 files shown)


zdrve:
I intend to do a similar job on 'generator' and 'saver' in forthcoming PRs.

lib/kennel.rb
```diff
@@ -127,10 +132,10 @@ def generated
           load_all
 
           projects = Models::Project.recursive_subclasses.map(&:new)
-          filter_resources!(projects, :kennel_id, project_filter, "projects", "PROJECT")
+          Kennel::Filter.filter_resources!(projects, :kennel_id, filter.project_filter, "projects", "PROJECT")
```
grosser, line 135: might make sense to:
```suggestion
          filter.filter_projects!(projects)
```
so the filter can hold have all the "PROJECT" logic

grosser: working on that ...

grosser: https://github.com/grosser/kennel/pull/192

PR #192, opened 2022-09-25
move more filtering logic into filter

grosser:
@zdrve

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -101,10 +101,10 @@ module Kennel
       @generated ||= begin
         Progress.progress "Generating" do
           projects = projects_provider.projects
-          Kennel::Filter.filter_resources!(projects, :kennel_id, filter.project_filter, "projects", "PROJECT")
+          projects = filter.filter_projects projects
 
           parts = Utils.parallel(projects, &:validated_parts).flatten(1)
-          Kennel::Filter.filter_resources!(parts, :tracking_id, filter.tracking_id_filter, "resources", "TRACKING_ID")
+          parts = filter.filter_parts parts
 
           parts.group_by(&:tracking_id).each do |tracking_id, same|
             next if same.size == 1
--- a/lib/kennel/filter.rb
+++ b/lib/kennel/filter.rb
@@ -2,35 +2,48 @@
 
 module Kennel
   class Filter
+    attr_reader :project_filter, :tracking_id_filter
+
     def initialize
-      project_filter
-      tracking_id_filter
+      # build early so we fail fast on invalid user input
+      @tracking_id_filter = build_tracking_id_filter
+      @project_filter = build_project_filter
     end
 
-    def project_filter
-      projects = ENV["PROJECT"]&.split(",")&.sort&.uniq
-      tracking_projects = tracking_id_filter&.map { |id| id.split(":", 2).first }&.sort&.uniq
-      if projects && tracking_projects && projects != tracking_projects
+    def filter_projects(projects)
+      filter_resources(projects, :kennel_id, project_filter, "projects", "PROJECT")
+    end
+
+    def filter_parts(parts)
+      filter_resources(parts, :tracking_id, tracking_id_filter, "resources", "TRACKING_ID")
+    end
+
+    private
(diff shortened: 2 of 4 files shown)


zdrve (changes requested):
Politely declining this since, as I said, I was making more PRs.

zdrve:
I will revisit this PR and see if it makes sense once #191 and #193 are in. :-)

grosser:
rebased ... looks good ?

lib/kennel.rb
```diff
@@ -101,10 +101,10 @@ def generated
       @generated ||= begin
         Progress.progress "Generating" do
           projects = projects_provider.projects
-          Kennel::Filter.filter_resources!(projects, :kennel_id, filter.project_filter, "projects", "PROJECT")
+          filter.filter_projects! projects
```
zdrve, line 104: Prefer `projects = filter.filter_projects(projects)`

test/kennel/filter_test.rb
```diff
@@ -4,61 +4,90 @@
...
+    context "with project, with tracking_id" do
+      it "works when they agree" do
+        f = with_env("PROJECT" => "foo,bar", "TRACKING_ID" => "foo:x,bar:y") do
           Kennel::Filter.new
         end
+        f.project_filter.must_equal(["bar", "foo"])
+        f.tracking_id_filter.must_equal(["bar:y", "foo:x"])
+      end
 
+      it "fails when they disagree" do
```
zdrve, line 37:
```suggestion
      it "raises when they disagree" do
```

test/kennel/filter_test.rb
```diff
@@ -4,61 +4,90 @@
...
+      parts = [stub("P1", tracking_id: "a"), stub("P2", tracking_id: "b")]
+      with_env TRACKING_ID: "a" do
+        Kennel::Filter.new.filter_parts!(parts)
       end
+      parts.map(&:tracking_id).must_equal ["a"]
     end
   end
 
-  describe ".filter_resources!" do
+  describe "#filter_resources!" do
```
zdrve, line 90: Do we really want to be testing a private method? That can be OK, but is worth a comment to explain what's going on, and why we're doing that.

grosser: - don't want to re-do the whole test
- it's kind of nice since the 2 public methods share it
... added a comment explaining

grosser:
updated 👁️

test/kennel/filter_test.rb
zdrve: Looks good.

This PR echoes work that I had already done, I remember writing the exact same code, and indeed even taking the same approach with testing via the private method – with a comment 😛 – but I didn't commit it / turn it into a PR because I chose to focus on refactoring the other areas first :-)

lib/kennel/filter.rb
```diff
@@ -2,32 +2,45 @@
...
     def initialize
-      project_filter
-      tracking_id_filter
+      # build early so we fail fast on invalid user input
+      @tracking_id_filter = build_tracking_id_filter
+      @project_filter = build_project_filter
+    end
+
+    def filter_projects(projects)
+      filter_resources!(projects.dup, :kennel_id, project_filter, "projects", "PROJECT")
```
zdrve, line 14: Let's keep the dup-ness as an implementation detail of the private method

```suggestion
      filter_resources(projects, :kennel_id, project_filter, "projects", "PROJECT")
```

lib/kennel/filter.rb
```diff
@@ -2,32 +2,45 @@
...
     end
 
-    def tracking_id_filter
+    def build_tracking_id_filter
       (tracking_id = ENV["TRACKING_ID"]) && tracking_id.split(",").sort.uniq
     end
 
-    def self.filter_resources!(resources, by, against, name, env)
-      return unless against
+    def filter_resources!(resources, by, against, name, env)
```
zdrve, line 36: To help readability. 

```suggestion
    def filter_resources!(resources, by, allow_list, name, env)
```

zdrve (approved):

PR #193, opened 2022-09-25
Split parts writer

zdrve:
Includes a bug fix, whereby old things in `generated` weren't cleaned up.
Applying this branch to zendesk kennel 082582d0425f16d2a314e90214c1b7e7c3c53867, this means that suddenly 56 things under "generated" get deleted, even though "rake plan" correctly shows "no changes".

Also clarifies that when `TRACKING_ID` filtering is used, cleanup is skipped.

Builds on #191

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -9,6 +9,7 @@ require "kennel/compatibility"
 require "kennel/utils"
 require "kennel/progress"
 require "kennel/filter"
+require "kennel/parts_writer"
 require "kennel/projects_generator"
 require "kennel/syncer"
 require "kennel/id_map"
@@ -57,7 +58,7 @@ module Kennel
 
     def generate
       out = generated
-      store out if ENV["STORE"] != "false" # quicker when debugging
+      parts_writer.write(out) if ENV["STORE"] != "false" # quicker when debugging
       out
     end
 
@@ -76,45 +77,6 @@ module Kennel
 
     private
 
-    def store(parts)
-      Progress.progress "Storing" do
-        old = Dir[[
-          "generated",
-          if filter.project_filter || filter.tracking_id_filter
-            [
-              "{" + (filter.project_filter || ["*"]).join(",") + "}",
-              "{" + (filter.tracking_id_filter || ["*"]).join(",") + "}.json"
-            ]
-          else
-            "**"
-          end
-        ].join("/")]
-        used = []
-
-        Utils.parallel(parts, max: 2) do |part|
-          path = "generated/#{part.tracking_id.tr("/", ":").sub(":", "/")}.json"
-          used.concat [File.dirname(path), path] # only 1 level of sub folders, so this is safe
-          payload = part.as_json.merge(api_resource: part.class.api_resource)
-          write_file_if_necessary(path, JSON.pretty_generate(payload) << "\n")
-        end
-
-        # deleting all is slow, so only delete the extras
-        (old - used).each { |p| FileUtils.rm_rf(p) }
-      end
-    end
-
-    def write_file_if_necessary(path, content)
-      # 99% case
-      begin
-        return if File.read(path) == content
-      rescue Errno::ENOENT
-        FileUtils.mkdir_p(File.dirname(path))
-      end
-
-      # slow 1% case
-      File.write(path, content)
-    end
-
     def filter
       @filter ||= Filter.new
     end
@@ -129,6 +91,10 @@ module Kennel
 
     def projects_generator
       @projects_generator ||= ProjectsGenerator.new
+    end
+
+    def parts_writer
+      @parts_writer ||= PartsWriter.new(filter: filter)
     end
 
     def generated
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -10,7 +10,7 @@ require "kennel/utils"
 require "kennel/progress"
 require "kennel/filter"
 require "kennel/parts_writer"
-require "kennel/projects_generator"
+require "kennel/projects_provider"
 require "kennel/syncer"
 require "kennel/id_map"
 require "kennel/api"
@@ -90,7 +90,7 @@ module Kennel
     end
 
     def projects_generator
-      @projects_generator ||= ProjectsGenerator.new
+      @projects_generator ||= ProjectsProvider.new
     end
 
     def parts_writer
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -57,9 +57,9 @@ module Kennel
     attr_accessor :out, :err, :strict_imports
 
     def generate
-      out = generated
-      parts_writer.write(out) if ENV["STORE"] != "false" # quicker when debugging
-      out
+      parts = generated
+      parts_writer.write(parts) if ENV["STORE"] != "false" # quicker when debugging
+      parts
     end
 
     def plan
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -9,6 +9,7 @@ require "kennel/compatibility"
 require "kennel/utils"
 require "kennel/progress"
 require "kennel/filter"
+require "kennel/projects_generator"
 require "kennel/syncer"
 require "kennel/id_map"
 require "kennel/api"
@@ -126,12 +127,14 @@ module Kennel
       @api ||= Api.new(ENV.fetch("DATADOG_APP_KEY"), ENV.fetch("DATADOG_API_KEY"))
     end
 
+    def projects_generator
+      @projects_generator ||= ProjectsGenerator.new
+    end
+
     def generated
       @generated ||= begin
         Progress.progress "Generating" do
-          load_all
-
-          projects = Models::Project.recursive_subclasses.map(&:new)
+          projects = projects_generator.projects
           Kennel::Filter.filter_resources!(projects, :kennel_id, filter.project_filter, "projects", "PROJECT")
 
           parts = Utils.parallel(projects, &:validated_parts).flatten(1)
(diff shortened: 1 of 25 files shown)


lib/kennel.rb
```diff
@@ -55,9 +57,9 @@ def initialize
     attr_accessor :out, :err, :strict_imports
 
     def generate
-      out = generated
-      store out if ENV["STORE"] != "false" # quicker when debugging
-      out
+      parts = generated
+      parts_writer.write(parts) if ENV["STORE"] != "false" # quicker when debugging
```
grosser, line 61: `serializer` might be a better name for parts-writer

lib/kennel.rb
```diff
@@ -126,12 +89,18 @@ def api
       @api ||= Api.new(ENV.fetch("DATADOG_APP_KEY"), ENV.fetch("DATADOG_API_KEY"))
     end
 
+    def projects_generator
+      @projects_generator ||= ProjectsGenerator.new
```
grosser, line 93: how about "loader" ?
```suggestion
      @projects_generator ||= ProjectsLoader.new
```

lib/kennel/parts_writer.rb
```diff
@@ -0,0 +1,70 @@
...
+      used = ["generated"]
+
+      Utils.parallel(parts, max: 2) do |part|
+        path = "generated/#{part.tracking_id.tr("/", ":").sub(":", "/")}.json"
+        used.concat [File.dirname(path), path] # only 1 level of sub folders, so this is safe
+        payload = part.as_json.merge(api_resource: part.class.api_resource)
+        write_file_if_necessary(path, JSON.pretty_generate(payload) << "\n")
+      end
+
+      used
```
grosser, line 37: since we can add the folder multiple times we should `uniq`
```suggestion
      used.uniq
```

lib/kennel/parts_writer.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def apply_cleanup_to
```
grosser, line 40: `apply_to` sounds like it's doing an action
```suggestion
    def directory_for_cleanup
```

lib/kennel/parts_writer.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def old_paths
+      apply_cleanup_to.flat_map do |path|
+        if File.exist?(path)
+          Find.find(path).to_a
+        else
+          []
+        end
+      end
+    end
```
grosser, line 56: finding all files adds overhead to the "generate all" usecase, so prefer to avoid

test/kennel/parts_writer_test.rb
```diff
@@ -0,0 +1,172 @@
...
+  let(:filter) do
+    filter = "some filter".dup
+    filter.stubs(:project_filter).returns(project_filter)
+    filter.stubs(:tracking_id_filter).returns(tracking_id_filter)
+    filter
+  end
+
+  capture_all
+
+  def make_project(kennel_id, monitor_kennel_ids)
```
grosser, line 26: prefer to have all lets / all defs / all setups + 'in_temp_dir / capture_all' in a single place and not interleaved

test/kennel/parts_writer_test.rb
```diff
@@ -0,0 +1,172 @@
...
+            critical: 123
+          )
+        end
+      }
+    )
+  end
+
+  in_temp_dir
+
+  it "saves formatted json" do
```
grosser, line 50: wrap everything in `describe "#write"` to make adding more methods/tests not require a big reformat

test/test_helper.rb
```diff
@@ -18,27 +18,31 @@
...
 require "kennel"
 
-class TestProject < Kennel::Models::Project
-  defaults(
-    team: -> { TestTeam.new },
-    parts: -> { [] }
-  )
-end
+Minitest::Test.class_eval do
+  def self.with_test_classes
```
grosser, line 22: should this not clean up the classes after too ?
```
def with_test_classes
  around do |t|
    classes
    t.call
  ensure
    delete classes
  end
end
```

grosser: also a bit sus that it calls when the test is loaded and not when the test is run
... might as well always create the classes in test_helper if we really need this, because otherwise we have different behavior depending on what test files are loaded

grosser (commented):
hard to review when it mixes moving code with refactoring code 😞

lib/kennel/parts_writer.rb
zdrve:
```
$ time ruby -rfind -rbenchmark -e 'puts Benchmark.measure { 100.times { Find.find("generated").to_a } }'
  5.003687   8.399515  13.403202 ( 13.436303)
ruby -rfind -rbenchmark -e   5,07s user 8,43s system 99% cpu 13,541 total
```

It seems to be inconsequential. Premature optimization and all that.

test/kennel/parts_writer_test.rb
zdrve: What do you mean? All the defs together, all the lets together, etc?

test/test_helper.rb
zdrve: The criticisms you make apply equally well to the pre-PR code too. In fact even more so, because the whole point of `with_test_classes` is to make that suspect behaviour optional.

lib/kennel/parts_writer.rb
zdrve: Sure. Again, this was a problem with the original code too, but I am happy that refactoring like this is helping bringing these problems to the fore so they can be fixed :)

lib/kennel.rb
zdrve: Well if anything it's more like a `Provider`. It's a thing which can provide a set of projects. This one just happens to do it by using Zeitwerk / load_all and then doing the all-subclasses-of-Project thing. The intention is that, for testing, it will be (i.e. in future PRs) useful to supply other ways of providing projects (e.g. "here is an array of projects"), because then that's much easier to work with for testing. You don't need temp dirs, you don't need to write to the filesystem, you don't need to unload classes and mess around with `Object.send(:remove_const)`.

But having said that: it's just a name, so we can call it `Loader` if you like and still do the testing stuff I just said.

lib/kennel.rb
zdrve: Maybe. I mean ... again (see ProjectLoader thread) it's just an internal name, so whatever.

lib/kennel/parts_writer.rb
grosser: so 100ms right ? ... prefer to avoid that
... also means lots of disk-io
I'm hoping that there is a cheaper way of doing this ... ideally in it's own PR so we can code-golf more
... if it's complicated to take back out then just leave it and I'll take a look once it's merged

test/kennel/parts_writer_test.rb
grosser: something like
```
def foo
def bar

let(:baz)
let(:fop)

in_tempdir
before { stuff }
after { stuff }
```

test/test_helper.rb
grosser: before everything had the same environment
after:
- the classes are defined multiple times
- depending on the load-order of the tests different classes are defined

lib/kennel.rb
zdrve: dd9883e62fcb84fe3f79879525d52c47161f5c61

lib/kennel/parts_writer.rb
zdrve: ecb1a4ab81983270ed6e012f63f0d490e279df6b

lib/kennel/parts_writer.rb
zdrve: 6cb0ed1a1eb2d0015596b5004f8bc4febcf013a2

test/kennel/parts_writer_test.rb
zdrve: d813cfb4905f515250f9131e9cc047d3ddac0081

test/kennel/parts_writer_test.rb
zdrve: 843c5ccfa1fb27302d2de788a7288ffbd5c273a2

lib/kennel.rb
zdrve: 6c8096b

lib/kennel/parts_writer.rb
zdrve: Given that the pre-PR version is buggy, let's concentrate on correctness first, and maybe optimize later.

test/test_helper.rb
zdrve: Repeating yourself can be fine. Can be a _good_ thing.

Could you be more specific about the repeat definitions / the load order thing?

grosser: when multiple tests are loaded the method gets called multiple times
so the classes get re-defined multiple times (right ?)

when multiple tests get loaded at the same time the defined classes will be different from when only a single test is loaded
... so having a single global env might be simpler and easier to debug

zdrve: > when multiple tests are loaded the method gets called multiple times 

Does that ever happen? AFAICT each test file runs in its own process.

> when multiple tests get loaded at the same time

See above.

grosser: ah true! :D
... yeah with forking-test-runner this should be fine 👍

grosser:
lmk when you think it's ready to go 👍

lib/kennel.rb
```diff
@@ -126,12 +89,18 @@ def api
       @api ||= Api.new(ENV.fetch("DATADOG_APP_KEY"), ENV.fetch("DATADOG_API_KEY"))
     end
 
+    def projects_generator
+      @projects_generator ||= ProjectsProvider.new
+    end
```
grosser, line 94:
```suggestion
    def projects_provider
      @projects_provider ||= ProjectsProvider.new
    end
```

zdrve: 079b3bafa99c6b2109b59220f30bdacb813be06a

zdrve:
Good to go 👍

PR #195, opened 2022-09-26
speedup cleanup by using Dir instead of Find

grosser:
```
>> Benchmark.realtime { Find.find("generated").to_a }
=> 0.11841299990192056
>> Benchmark.realtime { Dir["generated/**/*"] }
=> 0.040566000156104565
```

@zdrve

### DIFF
--- a/lib/kennel/parts_serializer.rb
+++ b/lib/kennel/parts_serializer.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require "find"
 require "set"
 
 module Kennel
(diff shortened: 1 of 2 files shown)


test/kennel/parts_serializer_test.rb
```diff
@@ -37,10 +36,11 @@ def make_project(kennel_id, monitor_kennel_ids)
   let(:tracking_id_filter) { nil }
 
   let(:filter) do
-    filter = "some filter".dup
-    filter.stubs(:project_filter).returns(project_filter)
-    filter.stubs(:tracking_id_filter).returns(tracking_id_filter)
-    filter
+    stub(
```
zdrve, line 39: TIL :-)

lib/kennel/parts_serializer.rb
```diff
@@ -48,13 +49,7 @@ def directories_to_clean_up
...
 
     def old_paths
-      directories_to_clean_up.flat_map do |path|
-        if File.exist?(path)
-          Find.find(path).to_a
-        else
-          []
-        end
-      end.to_set
+      Dir["{#{directories_to_clean_up.join(",")}}/**/*"].to_set
```
zdrve, line 52: Edge case that this implementation introduces, FYI: PROJECT=foo, where a project with kennel_id=foo exists, but has no parts. Pre-PR: `generated/foo` would be deleted. Post-PR, it wouldn't.

zdrve (approved):
Seems like an unnecessary optimisation to me, but if it's important to you, go for it :-)

PR #197, opened 2022-09-27
enforce that composite monitors use links

grosser:


zdrve:
1. It's very much a non-backwards-compatible change (mainly of interest to anyone using Kennel who isn't Zendesk)
2. If, as someone using kennel outside of Zendesk, you wanted to turn _off_ this behaviour that this PR adds – i.e. you're fine with link-by-datadog-id, without the need to opt out each time – how would you do that?
3. The only way to opt out is to opt out of _all_ validations for the monitor.
4. How is the mutable overridable array meant to be used? Can you sketch what that would look like for Zendesk's case?

grosser:
1: it requires some code changes, but not too bad I'd hope
2: override the method in the monitor class to be noop
3: it's a setting so link validation can be turned off on it's own
4: `ALLOWED_UNLINKED = YAML.load_file('our_allow_list.yam')` or hard-coded list in the code

lib/kennel/models/monitor.rb
```diff
@@ -238,6 +239,8 @@ def validate_json(data)
           validate_message_variables(data)
         end
 
+        validate_using_links(data)
```
zdrve, line 242: Prefer a boolean setting:

```suggestion
        validate_using_links(data) if require_using_links
```

Having a "setting" where the two values are "method with logic" or "override with a no-op block" is ... a weird way of programming.

zdrve: Particularly when the method/block in question effectively returns void

lib/kennel/models/monitor.rb
```diff
@@ -27,11 +27,12 @@ class Monitor < Record
       }.freeze
       DEFAULT_ESCALATION_MESSAGE = ["", nil].freeze
       ALLOWED_PRIORITY_CLASSES = [NilClass, Integer].freeze
+      ALLOWED_UNLINKED = [] # rubocop:disable Style/MutableConstant placeholder for custom overrides
```
zdrve, line 30: If (when) we extend this validation to other linking types - e.g. monitor widgets in dashboards – what will ALLOWED_UNLINKED look like then?

grosser: will need to move up into `Kennel::ALLOWED_UNLINKED` but the structure can stay the same

lib/kennel/models/monitor.rb
grosser: yeah it's not very smooth ... some more general `skip_validations: [:foo]` might be nicer ... but that's more refactoring and can be it's own PR

grosser: another option would be
```
eval "->(*) { v }", nil, file, line.to_i
```
to use `validate_using_links: false`
but that ends up making any setting callable which could end up ugly too

PR #199, opened 2022-09-30
allow using escaped replacement style

grosser:
dd requires use of
```
{{[http.status_code].name}}
```
to be able to use a variable with a . in a message, so we should support it too


dyanawu (approved):
Amazing, thank you <3

zdrve (approved):

PR #201, opened 2022-10-01
working_json

zdrve:
A solution for the "static parts" problem.

 * `as_json` is now (normally) called exactly once per part for each run (therefore there's no need to memoize this method)
 * `working_json` is a kennel-internal method; it returns the memoized "working copy" of the json, defaulting to `as_json`. "Working copy" refers to the fact that the json mutates during processing.
 * `working_json!` can be used to reset the memo of `working_json`.

This means that even if a project returns a "static" part, `Kennel::Engine.new.plan` can be called more than once.

### DIFF
--- a/lib/kennel/deep_freeze.rb
+++ b/lib/kennel/deep_freeze.rb
@@ -1,0 +1,28 @@
+module Kennel
+  module DeepFreeze
+    # There'll be a gem for this somewhere.
+    # This code doesn't handle cycles or other reused references.
+
+    def deep_freeze(item)
+      case item
+      when Hash
+        item.map { |k, v| [ deep_freeze(k), deep_freeze(v) ] }.to_h
+      when Array
+        item.map { |v| deep_freeze(v) }
+      else
+        item.dup.freeze
+      end.freeze
+    end
+
+    def deep_dup_thaw(value)
+      case value
+      when Array
+        value.map { |v| deep_dup_thaw(v) }
+      when Hash
+        value.map { |k, v| [ deep_dup_thaw(k), deep_dup_thaw(v) ] }.to_h
+      else
+        value.dup
+      end
+    end
+  end
+end
--- a/lib/kennel/deep_freeze.rb
+++ b/lib/kennel/deep_freeze.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Kennel
   module DeepFreeze
     # There'll be a gem for this somewhere.
@@ -6,7 +8,7 @@ module Kennel
     def deep_freeze(item)
       case item
       when Hash
-        item.map { |k, v| [ deep_freeze(k), deep_freeze(v) ] }.to_h
+        item.map { |k, v| [deep_freeze(k), deep_freeze(v)] }.to_h
       when Array
         item.map { |v| deep_freeze(v) }
       else
@@ -19,7 +21,7 @@ module Kennel
       when Array
         value.map { |v| deep_dup_thaw(v) }
       when Hash
-        value.map { |k, v| [ deep_dup_thaw(k), deep_dup_thaw(v) ] }.to_h
+        value.map { |k, v| [deep_dup_thaw(k), deep_dup_thaw(v)] }.to_h
       else
         value.dup
       end
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -81,7 +81,7 @@ module Kennel
       end
 
       def diff(actual)
-        expected = as_json
+        expected = working_json
         expected.delete(:id)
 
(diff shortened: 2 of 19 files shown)


grosser (commented):
prefer to not add more runtime overhead / complexity for the notification feature 😞

zdrve:
I know you would. The difference is not statistically significant. Please request a specific change, or approve.

grosser:
how is this better than doing `parts.each { |p| p.instance_variable_set :@json, nil }` (or a nice method that wraps this) ?

zdrve:
`working_json!` is exactly the nice wrapper method which you suggest. The rest of the PR is:

 * introducing "working_json" as a name to highlight the mutating nature of the data therein
 * adding a deep clone, so as to provide stronger guarantees about the data we're working with, to make the code easier to reason about

grosser:
so benchmarking shows something strange:
master:
```
Generating ... 7.96s
Storing ... 2.11s
run  10: 11.02
```

this branch:
```
Generating ... 8.55s
Storing ... 1.01s
run  10: 10.31
```

... somehow it made generation slower but storing faster ?
the thaw would be an obvious culprit for the slowness, but why is storage faster ?

lib/kennel/deep_freeze.rb
```diff
@@ -0,0 +1,30 @@
...
+
+module Kennel
+  module DeepFreeze
+    # There'll be a gem for this somewhere.
+    # This code doesn't handle cycles or other reused references.
+
+    def deep_freeze(item)
+      case item
+      when Hash
+        item.map { |k, v| [deep_freeze(k), deep_freeze(v)] }.to_h
```
grosser, line 11:
```suggestion
        item.to_h { |k, v| [deep_freeze(k), deep_freeze(v)] }
```

lib/kennel/deep_freeze.rb
```diff
@@ -0,0 +1,30 @@
...
+        item.dup.freeze
+      end.freeze
+    end
+
+    def deep_dup_thaw(value)
+      case value
+      when Array
+        value.map { |v| deep_dup_thaw(v) }
+      when Hash
+        value.map { |k, v| [deep_dup_thaw(k), deep_dup_thaw(v)] }.to_h
```
grosser, line 24:
```suggestion
        value.to_h { |k, v| [deep_dup_thaw(k), deep_dup_thaw(v)] }
```

lib/kennel/models/record.rb
```diff
@@ -80,8 +82,17 @@ def initialize(project, *args)
         super(*args)
       end
 
+      def working_json
```
grosser, line 85: some docs here on why we are doing that would be good

lib/kennel/deep_freeze.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Kennel
+  module DeepFreeze
+    # There'll be a gem for this somewhere.
+    # This code doesn't handle cycles or other reused references.
+
+    def deep_freeze(item)
```
grosser, line 8: never used ?

lib/kennel/deep_freeze.rb
```diff
@@ -0,0 +1,30 @@
...
+      when Hash
+        item.map { |k, v| [deep_freeze(k), deep_freeze(v)] }.to_h
+      when Array
+        item.map { |v| deep_freeze(v) }
+      else
+        item.dup.freeze
+      end.freeze
+    end
+
+    def deep_dup_thaw(value)
```
grosser, line 19: prefer be `self.` and used statically instead

grosser:
looks like we can use Marshal instead:
```
>> Benchmark.realtime { 1000.times { Marshal.load Marshal.dump(a: [{b: {c: 3}}]) } }
=> 0.004929000046104193
>> Benchmark.realtime { 1000.times { deep_dup_thaw a: [{b: {c: 3}}] } }
=> 0.006802999880164862
```

grosser:
... for performance we can turn the duping on only when doing multiple passes ?

zdrve:
> ... for performance we can turn the duping on only when doing multiple passes ?

Let's not; sounds like a very poor complexity / payoff ratio.

All of your other feedback is 👍

zdrve:
>  the thaw would be an obvious culprit for the slowness, but why is storage faster ?

This reminds me, there was something odd I wanted to investigate. I don't have an answer for your question but thank you for the unintended prompt :-)

grosser:
if we simplify it down to Marshal then I'd suggest moving it into `deep_dup` in utils.rb
https://github.com/grosser/kennel/blob/43d5e1c2d4df68a80b71025ea5ab165472e61f53/test/test_helper.rb#L88-L90

PR #202, opened 2022-10-02
Syncer refactor

zdrve:
Including:

* adding a FIXME comment for a significant gap in testing
* instead of adding tracking_id and then removing it again, we never add it in the first place
* use accessors more

Also remove the 0.2 second minimum duration in `Progress.progress`. Speeds up the syncer tests hugely.


grosser:
not adding tracking_id and the min removal sounds interesting, but would prefer as standalone PR

zdrve:
Became #234

PR #203, opened 2022-10-03
Speed up tests by eliminating mandatory 0.2 second wait

zdrve:
Will also speed up "production" usage of `Progress.progress`,  but since it's an average of 0.1 seconds per invocation, it won't be much there, proportionally.


grosser (commented):
I can't tell if this is correct in all cases, will take another look later.
The "wake up when done" is nice though.

zdrve:
:-(

PR #204, opened 2022-10-04
make progress stop when block stops

grosser:
replaces https://github.com/grosser/kennel/pull/203

@zdrve

### DIFF
--- a/lib/kennel/progress.rb
+++ b/lib/kennel/progress.rb
@@ -4,7 +4,7 @@ require "benchmark"
 module Kennel
   class Progress
     # print what we are doing and a spinner until it is done ... then show how long it took
-    def self.progress(name)
+    def self.progress(name, interval: 0.2)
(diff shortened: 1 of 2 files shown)


zdrve:
It's not as watertight as my PR, but it'll help most of the time, I'll give you that.

grosser:
I'm not aware of any edge-case or bug that is not addressed, can you elaborate ?

zdrve:
```
         break if stop
                         # <- the other thread calls 'thread.run' here
          Kennel.err.print animation[count % animation.size]
                         # <- or here
          sleep interval
```

in which case the `.run` won't wake up the thread, because it hasn't slept yet.

grosser:
yeah true, good catch :)
0.0x% chance that we lose 0.2s so should be fine :)

grosser:
found 1 more race condition that could crash the progress https://github.com/grosser/kennel/pull/205

PR #206, opened 2022-10-08
resolve composite monitor links when importing

grosser:
before raw ids and a validation failure when used:
```
  query: -> { "111 && 222 && 333" },
```

after resoled ids if they are known kennel monitors:
```
  query: -> { "%{a:b} && %{b:c} && 333" },
```

decided not to auto-add validation skips in the hope that whoever runs the import will manually fix them up

/cc @zdrve

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -406,6 +406,14 @@ https://foo.datadog.com/monitor/123
 ### Find all monitors with No-Data
 `rake kennel:nodata TAG=team:foo`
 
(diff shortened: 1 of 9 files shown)


zdrve:
We can extend this (perhaps in later PRs) to the other link types, right?

zdrve (approved):

grosser:
yeah, ideally all types would do this
... and it would be inside monitor etc
but easy to move 🤞

PR #207, opened 2022-10-13
Make better use of Kennel.{out,err}.tty?

zdrve:
To avoid finding the odd "\\" and "/" and so on (spinner debris) when redirecting the outputs.

Also removes the need for "strip_shell_control".

### DIFF
--- a/lib/kennel/github_reporter.rb
+++ b/lib/kennel/github_reporter.rb
@@ -24,7 +24,7 @@ module Kennel
     end
 
     def report(&block)
-      output = Utils.strip_shell_control(Utils.tee_output(&block).strip)
+      output = Utils.tee_output(&block).strip
     rescue StandardError
       output = "Error:\n#{$ERROR_INFO.message}"
       raise
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -42,7 +42,7 @@ module Kennel
 
     def confirm
       return false if noop?
-      return true if ENV["CI"] || !STDIN.tty?
+      return true if ENV["CI"] || !STDIN.tty? || !Kennel.err.tty?
       Utils.ask("Execute Plan ?")
     end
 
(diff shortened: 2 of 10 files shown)


lib/kennel/github_reporter.rb
```diff
@@ -24,7 +24,7 @@ def initialize(token, ref: "HEAD")
     end
 
     def report(&block)
-      output = Utils.strip_shell_control(Utils.tee_output(&block).strip)
```
grosser, line 27: this is removed because it won't have shell-control anymore ?

lib/kennel/syncer.rb
```diff
@@ -42,7 +42,7 @@ def plan
 
     def confirm
       return false if noop?
-      return true if ENV["CI"] || !STDIN.tty?
+      return true if ENV["CI"] || !STDIN.tty? || !Kennel.err.tty?
```
grosser, line 45: is checking only err for example good enough ? (just so all the checks are consistent)

lib/kennel/utils.rb
```diff
@@ -51,12 +51,10 @@ def ask(question)
         end
       end
 
-      def color(color, text)
-        "\e[#{COLORS.fetch(color)}m#{text}\e[0m"
-      end
+      def color(color, text, force: false)
+        return text unless force || Kennel.out.tty?
```
grosser, line 55: prefer this but 🤷 
```suggestion
        return text if !force && !Kennel.out.tty?
```

grosser (approved):
not adding and removing seems like a more efficient approach :D

lib/kennel/github_reporter.rb
zdrve: Yup, `tee_output` provides a non-tty, so won't have any escape sequences to remove.

lib/kennel/syncer.rb
zdrve: I suspect we don't need the ENV CI check, because I'd be surprised if the CI jobs run with a tty.

The err check is there because if you run the task with `< /dev/tty 2> err.log` then you'll never see the prompt.

grosser:
1.122.0

PR #208, opened 2022-10-13
Protect against cache corruption

zdrve:
I still occasionally see cache corruption:

```
$ with-ruby - bundle exec rake generate kennel:update_datadog PROJECT=guide-redirect-service
Generating ... 2.05s
Storing ... 0.0s
Downloading definitions ... 6.6s
Diffing ... -rake aborted!
NoMethodError: undefined method `reject!' for nil:NilClass
/Users/revans/git/github.com/grosser/kennel/lib/kennel/file_cache.rb:60:in `expire_old_data'
/Users/revans/git/github.com/grosser/kennel/lib/kennel/file_cache.rb:19:in `open'
...

$ mv tmp/cache tmp/cache.bad
$ with-ruby - bundle exec rake generate kennel:update_datadog PROJECT=guide-redirect-service
<works fine>
```

In this case, somehow a `nil` had been persisted. Not sure how, but here's a defence: 23e3842

Another form of possible corruption is partial writes. Fix: 75ff2dec0eed79f492cf8394e2f7c4ad05350e6c

### DIFF
--- a/lib/kennel/file_cache.rb
+++ b/lib/kennel/file_cache.rb
@@ -49,6 +49,7 @@ module Kennel
 
       Tempfile.create "kennel-file-cache", dir do |tmp|
         Marshal.dump @data, tmp
+        tmp.flush
         File.rename tmp.path, @file
       end
     end
--- a/lib/kennel/file_cache.rb
+++ b/lib/kennel/file_cache.rb
@@ -15,11 +15,13 @@ module Kennel
     end
 
     def open
-      load_data
-      expire_old_data
-      yield self
-    ensure
-      persist
+      @data = load_data || {}
+      begin
+        expire_old_data
+        yield self
+      ensure
+        persist
+      end
     end
 
     def fetch(key, key_version)
@@ -35,12 +37,9 @@ module Kennel
     private
 
     def load_data
-      @data =
-        begin
-          Marshal.load(File.read(@file)) # rubocop:disable Security/MarshalLoad
-        rescue StandardError
-          {}
-        end
+      Marshal.load(File.read(@file)) # rubocop:disable Security/MarshalLoad
+    rescue Errno::ENOENT, TypeError, ArgumentError
+      nil
     end
 
     def persist


lib/kennel/file_cache.rb
```diff
@@ -44,11 +44,14 @@ def load_data
...
 
     def persist
+      return if @data.nil?
+
       dir = File.dirname(@file)
       FileUtils.mkdir_p(dir) unless File.directory?(dir)
 
       Tempfile.create "kennel-file-cache", dir do |tmp|
         Marshal.dump @data, tmp
+        tmp.flush
```
grosser, line 54: that's the good part I think 🤞

grosser: ... actually `.close` would be better

lib/kennel/file_cache.rb
```diff
@@ -44,11 +44,14 @@ def load_data
     end
 
     def persist
+      return if @data.nil?
```
grosser, line 47: prefer to make load_data fallback to `{}` we should never have nil data here

lib/kennel/file_cache.rb
zdrve: No difference really, it's going to close anyway

lib/kennel/file_cache.rb
zdrve: OK, do have a specific patch to propose?

zdrve: I have a _suspicion_ as to how my cache ended up containing simply a `nil` (the cache file was literally 3 bytes long, the exact same 3 bytes that `Marshal.dump(nil)` gives), i.e. in that specific case, it wasn't the non-flushing problem, it was the nil problem. However ... we kinda _already_ fall back to `{}`, right? Except it doesn't seem to work.

grosser: on line 40 we should say `|| {}`

zdrve: That's kinda missing the point, isn't it? That says: if the cache is `nil` (i.e. already corrupted), then try to un-corrupt it. Shouldn't we be trying to make sure it doesn't corrupt in the first place?

grosser:
Yeah raise when it's about to store a nil so we can find the root cause 👍

zdrve:
How would that help find the root cause? 

This is an interesting discussion but I'm not finding it very "actionable" when it comes to this PR.

lib/kennel/file_cache.rb
```diff
@@ -15,11 +15,13 @@ def initialize(file, cache_version)
     end
 
     def open
-      load_data
-      expire_old_data
-      yield self
-    ensure
-      persist
```
grosser, line 22: I bet that's it 🎉

lib/kennel/file_cache.rb
grosser: close is more final than flush, I doubt it makes a difference, but we want to say "ok we are done with this file"

grosser:
yeah I bet the ensure block was the issue ... it crashes and then writes nil 🎉

PR #210, opened 2022-10-14
Clean up validation

zdrve:
Simple enough refactor for now, but this is leading to something bigger.

What *this* PR achieves is to have `ValidationError` refer solely to the type of error which one can get from the question: "Is this record, without visibility of any other records, and without visibility of Datadog, valid?".

Also minor rewording of the error messages of what is now UnresolvableIdError.

What this is hopefully leading to (later PRs):

  * make the record validation more consistent (for example, it makes no sense that the check in `OptionalValidations` applies to only two of the record types)
  * move `validate_json if validate` to outside of `#as_json` (it should be applied to _the result of_ `as_json`)
  * Give each kind of validation error a tag (e.g. `:no_data_timeframe_too_small`); deprecate `validate: false` (which suppresses _all_ validations for that record), and replace it by a more fine-grained mechism, `skip_validations: [ :no_data_timeframe_too_small ]`.

### DIFF
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -104,11 +104,11 @@ module Kennel
     def resolved?(e)
       assert_resolved e
       true
-    rescue ValidationError
+    rescue UnresolvableIdError
       false
     end
 
-    # raises ValidationError when not resolved
+    # raises UnresolvableIdError when not resolved
     def assert_resolved(e)
       resolve_linked_tracking_ids! [e], force: true
     end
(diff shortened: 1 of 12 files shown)


lib/kennel/syncer.rb
```diff
@@ -104,11 +104,11 @@ def each_resolved(list)
     def resolved?(e)
       assert_resolved e
       true
-    rescue ValidationError
+    rescue UnresolvableIdError
```
grosser, line 107: yeah that's nice! :)

PR #211, opened 2022-10-14
Make validation more consistent

zdrve:
Follow-up to #210.

 * make the record validation more consistent (for example, it makes no sense that the check in OptionalValidations applies to only two of the record types)
 * move validate_json if validate to outside of #as_json (it should be applied to the result of as_json)
 * use `ValidationError` exclusively for errors from `#validate_json`

**Note that this is a breaking change**. Any override of `as_json` should be changed to `build_json`, and any memoization in those overrides should be removed.

Still to follow in a later PR:

* Give each kind of validation error a tag (e.g. `:no_data_timeframe_too_small`); deprecate `validate: false` (which suppresses _all_ validations for that record), and replace it by a more fine-grained mechanism, e.g. `skip_validations: [ :no_data_timeframe_too_small ]`

### DIFF
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -30,6 +30,8 @@ module Kennel
 
       settings :id, :kennel_id
 
+      defaults(id: nil)
+
       class << self
         def parse_any_url(url)
           subclasses.detect do |s|
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -132,6 +132,7 @@ module Kennel
       def as_json
         @as_json ||= begin
                        json = build_json
+                       (id = json.delete(:id)) && json[:id] = id
                        validate_json(json) if validate
                        json
                      end
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -119,6 +119,16 @@ module Kennel
         self.class.remove_tracking_id(as_json)
       end
 
+      def build_json
+        {
+          id: id
+        }.compact
+      end
+
+      def as_json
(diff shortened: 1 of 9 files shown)


grosser:
rebase plz :)

lib/kennel/models/record.rb
```diff
@@ -96,7 +100,7 @@ def tracking_id
         @tracking_id ||= begin
           id = "#{project.kennel_id}:#{kennel_id}"
           unless id.match?(ALLOWED_KENNEL_ID_REGEX) # <-> parse_tracking_id
-            raise ValidationError, "#{id} must match #{ALLOWED_KENNEL_ID_REGEX}"
+            raise "#{id} must match #{ALLOWED_KENNEL_ID_REGEX}"
```
grosser, line 103: would be nice to have a type for this too (instead of RuntimeError)

lib/kennel/models/record.rb
```diff
@@ -119,6 +123,21 @@ def remove_tracking_id
...
+        }.compact
+      end
+
+      def as_json
+        @as_json ||= begin
+                       json = build_json
+                       (id = json.delete(:id)) && json[:id] = id
+                       validate_json(json) if validate
+                       json
+                     end
```
grosser, line 138: consistent indent is nice :)
```suggestion
        @as_json ||=
          begin
            json = build_json
            (id = json.delete(:id)) && json[:id] = id
            validate_json(json) if validate
            json
          end
```

grosser (approved):

grosser:
I like the consistency! :)

PR #212, opened 2022-10-15
Validation error tagging

zdrve:
Builds on #211 (and includes all of its commits).

Give validation errors tags, e.g. `:query_must_compare_against_critical`. Prefer suppression of individual errors (`skip_validations: [:query_must_compare_against_critical]`) over blanket suppression of all errors (`validate: false`). It is an error to use validation-suppression when no validation errors in fact occur.

The last commit is especially messy, and introduces a little non-test-covered code; I got lazy. A lot of the complexity in the last commit can go away once zendesk-kennel has migrated from `validate` to `skip_validations`.

### DIFF
--- a/lib/kennel/models/dashboard.rb
+++ b/lib/kennel/models/dashboard.rb
@@ -230,7 +230,7 @@ module Kennel
 
         # Avoid diff from datadog presets sorting.
         presets = data[:template_variable_presets]
-        invalid! :xxx1, "template_variable_presets must be sorted by name" if presets && presets != presets.sort_by { |p| p[:name] }
+        invalid! :template_variable_presets_must_be_sorted, "template_variable_presets must be sorted by name" if presets && presets != presets.sort_by { |p| p[:name] }
       end
 
       def render_definitions(definitions)
--- a/lib/kennel/models/dashboard.rb
+++ b/lib/kennel/models/dashboard.rb
@@ -230,7 +230,7 @@ module Kennel
 
         # Avoid diff from datadog presets sorting.
         presets = data[:template_variable_presets]
-        invalid! "template_variable_presets must be sorted by name" if presets && presets != presets.sort_by { |p| p[:name] }
+        invalid! :xxx1, "template_variable_presets must be sorted by name" if presets && presets != presets.sort_by { |p| p[:name] }
       end
 
       def render_definitions(definitions)
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -133,7 +133,7 @@ module Kennel
         @as_json ||= begin
                        json = build_json
                        (id = json.delete(:id)) && json[:id] = id
-                       validate_json(json) if validate
+                       validate_json_respecting_filters(json)
                        json
                      end
       end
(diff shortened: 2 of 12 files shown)


grosser:
1 rebase plz :)

zdrve:
Setting to draft because I want to have the error messages give more accurate feedback

lib/kennel/models/dashboard.rb
```diff
@@ -230,7 +230,7 @@ def validate_json(data)
 
         # Avoid diff from datadog presets sorting.
         presets = data[:template_variable_presets]
-        invalid! "template_variable_presets must be sorted by name" if presets && presets != presets.sort_by { |p| p[:name] }
+        invalid! :template_variable_presets_must_be_sorted, "template_variable_presets must be sorted by name" if presets && presets != presets.sort_by { |p| p[:name] }
```
grosser, line 233: could shorten that a bit:
```suggestion
        invalid! :template_variable_presets_sorted, "template_variable_presets must be sorted by name" if presets && presets != presets.sort_by { |p| p[:name] }
```

lib/kennel/models/record.rb
```diff
@@ -148,6 +148,52 @@ def invalid_update!(field, old_value, new_value)
...
+        caught = begin
+                   validate_json(data)
+                   nil
+                 rescue ValidationError => e
+                   e
+                 end
+
+        if caught.nil?
+          unless skip_validations.empty?
+            invalid! nil, "skip_validations must not be used when no validations need skipping"
```
grosser, line 161: this could backfire when used on a base monitor that others inherit from

lib/kennel/models/record.rb
```diff
@@ -148,6 +148,52 @@ def invalid_update!(field, old_value, new_value)
...
+            end
+
+            return
+          end
+
+          if skip_validations.include?(caught.tag)
+            # Ideally we'd check that skip_validations contains _only_ those
+            # validations while fail. However because the validation errors
+            # are raised exceptions, we can only see up to 1 per record, so
+            # we can't do that.
```
grosser, line 190: we could mark "used skips" and then compare against that
... but ideally we don't raise exceptions and then discard them since that's pretty slow, but make the validation calls do nothing if they are skipped (and mark "skip used" right there too)

lib/kennel/models/record.rb
```diff
@@ -179,8 +225,14 @@ def resolve_link(sought_tracking_id, sought_type, id_map, force:)
       end
 
       # let users know which project/resource failed when something happens during diffing where the backtrace is hidden
-      def invalid!(message)
-        raise ValidationError, "#{tracking_id} #{message}"
+      def invalid!(tag, message = nil)
+        if message.nil?
```
grosser, line 229: maybe some todo comment here to clean this up later

lib/kennel/models/record.rb
```diff
@@ -179,8 +225,14 @@ def resolve_link(sought_tracking_id, sought_type, id_map, force:)
...
+          message = tag
+          tag = :default_error_tag
+          Kennel.out.puts "I: untagged validation error '#{message}'" if ENV["SHOW_DISABLED_VALIDATION"]
```
grosser, line 232: maybe just leave it in the system as untagged and we can use some hacky code to find all the untagged places later

lib/kennel/models/record.rb
zdrve: Yeah I'm rewriting this whole section. 

Essentially the concept which I'm aiming for, setting _specific_ validations to be skipped rather than all or none, is only of any value if a failing (but skipped) validation is not terminal.

In other words, given a record which has problems with both foo and bar, then

* report error foo
* report error bar

doesn't works as intended if "report error" is terminal (i.e. `raise`) – because if foo raised, then we found out that foo was skipped, then we never got to see bar.

So I'm currently working on changing the error reporting to be non-terminal (within each record), i.e. we run as many validations as possible then process them, including skipping, once all errors have been collected from all parts. This also has the arguable benefit of reporting all the available problems at once, instead of trickle-feeding (problem / fix it / different problem shows up / fix it / ... ).

lib/kennel/models/record.rb
zdrve: In my current sketch of the branch, there are 7 possible ways that a message can be reported; it's quite a complex set of `if`s. Once we've completed the migration from `invalid` to `skip_validations`, it would be reduced to 4.

lib/kennel/models/record.rb
zdrve: I'm opting now for not having the "default tag" concept. On upgrade, overriders (e.g. zendesk-kennel) will have to decide whether each error is either skippable (in which case give it a tag), or not (in which case add the tag as `nil`).

PR #213, opened 2022-10-21
No raise on validation error

zdrve:
Adds `Record#build`.

`build` collects all validation errors, instead of stopping on the first one. `as_json` calls `build`, and raises the first validation error if any were encountered (i.e. the old behaviour).

We then use `build` inside `Kennel::Engine#generated`, instead of as_json. We report all validation errors of all records, instead of just the first one encountered.

Later on we can make the `validate` boolean setting more fine-grained.


lib/kennel/models/record.rb
```diff
@@ -129,25 +135,53 @@ def build_json
         }.compact
       end
 
+      # One of three outcomes:
+      # - throws (on non-validation error)
+      # - returns, validation_errors is not empty, and as_json is nil (invalid)
+      # - returns, validation_errors is empty, and as_json is not nil (valid)
+      def prepare
```
grosser, line 142: less generic name would be nice
prepare_validation ?

lib/kennel/models/record.rb
```diff
@@ -129,25 +135,53 @@ def build_json
         }.compact
       end
 
+      # One of three outcomes:
+      # - throws (on non-validation error)
```
grosser, line 139:
```suggestion
      # - raises (on non-validation error)
```

lib/kennel/models/record.rb
```diff
@@ -129,25 +135,53 @@ def build_json
...
+      # - returns, validation_errors is not empty, and as_json is nil (invalid)
+      # - returns, validation_errors is empty, and as_json is not nil (valid)
```
grosser, line 141: seems to repeat the code, maybe replace with a tldr

lib/kennel/models/record.rb
```diff
@@ -129,25 +135,53 @@ def build_json
...
+      def prepare
+        @validation_errors = []
+        json = nil
+
+        begin
+          json = build_json
+          (id = json.delete(:id)) && json[:id] = id
+          validate_json(json) if validate
+        rescue StandardError
```
grosser, line 150: why do we need to rescue here ? if we let it raise we get a nice backtrace too ?

lib/kennel/models/record.rb
```diff
@@ -129,25 +135,53 @@ def build_json
...
+          if validation_errors.empty?
+            @validation_errors = nil
+            raise PrepareError, safe_tracking_id
+          end
+        end
+
+        @as_json = (json if validation_errors.empty?)
```
grosser, line 157: can we always assign to have more info when debugging ?

lib/kennel/models/record.rb
```diff
@@ -129,25 +135,53 @@ def build_json
...
       def as_json
-        @as_json ||= begin
-                       json = build_json
-                       (id = json.delete(:id)) && json[:id] = id
-                       validate_json(json) if validate
-                       json
-                     end
+        # A courtesy to those tests that still expect as_json to perform validation and raise on error
+        prepare if @validation_errors.nil?
+        raise Kennel::ValidationError, "#{safe_tracking_id} #{@validation_errors.first}" unless validation_errors.empty?
```
grosser, line 163: should show all since this is what users will see ?

lib/kennel/models/record.rb
```diff
@@ -129,25 +135,53 @@ def build_json
...
+        raise DisallowedUpdateError, "#{safe_tracking_id} Datadog does not allow update of #{field} (#{old_value.inspect} -> #{new_value.inspect})"
       end
 
       private
 
+      # For use during error handling
+      def safe_tracking_id
+        tracking_id
+      rescue StandardError
+        "<unknown; #tracking_id crashed>"
```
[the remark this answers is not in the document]
zdrve: That's because you know only ever to set valid tracking ids :-)

lib/kennel/models/record.rb
zdrve: One of the things I'd love to do at some point would be to split the model class instance methods into two: pre-prepare and post-prepare. 

- pre-prepare has settings, and knows how to build_json / validate_json / prepare
- post-prepare has things like add_tracking_id, diff, resolve.

So conceptually calling the code ends up looking like:

```
monitor = K::M::Monitor.new(....) { ... }
built = monitor.build
built.as_json
built.add_tracking_id
# etc 
```

At which point "prepare" is basically the same as / replaced by "build".

For now, we could still call it `build` or `build_and_validate` if you like.

lib/kennel/models/record.rb
zdrve: Not sure what you mean. Are you saying that the comment seems superfluous, because the code is easy enough to understand as it is? Or something else?

lib/kennel/models/record.rb
zdrve: The thinking here is that because `invalid!` used to raise, no further processing happened beyond that. But that's not true any more; `invalid` sets state, and returns, potentially to allow more validation code to run. Since we've already encountered an error, we enter a kind of "all bets are off" state. For example, one might have:

```
def validate_json(data)
  data.key?(:tags) or invalid! "Need tags"

  data[:tags].all? { ... } or invalid! "..."
end

i.e. as soon as we've hit one validation error, we shouldn't be surprised if we crash beyond that.

This is what the rescue is for: to swallow errors _if_ they happen after we've already seen a validation error.

Comment (zdrve) [reply lib/kennel/models/record.rb:150]: Re. backtrace: at the moment on this branch, the output looks a bit like this:

```
Finding applicable projects and parts ... 2.11s
Preparing json ... 0.03s

user-segment-service:14a91074 type 'metric alert' is deprecated, please set to a different type (e.g. 'query alert')
user-segment-service:14a91074 priority needs to be an Integer
user-segment-service:user-segment-service-k8s-pods critical and value used in query must match

rake aborted!
Kennel::GenerationAbortedError: Kennel::GenerationAbortedError
<big useless backtrace>
```

i.e. each error is a single line, with the tracking id, and no backtrace. Do you want something different?

File: lib/kennel/models/record.rb (hunk 6 above)
Comment (zdrve) [reply lib/kennel/models/record.rb:163]: See my just-added comment further up.

File: lib/kennel/models/record.rb (hunk 1 above)
Comment (zdrve) [reply lib/kennel/models/record.rb:142]: I've called it `build`.

File: lib/kennel/models/record.rb (hunk 3 above)
Comment (grosser) [reply lib/kennel/models/record.rb:141]: to understand the comment I had to read the code :D

File: lib/kennel/models/record.rb (hunk 4 above)
Comment (grosser) [reply lib/kennel/models/record.rb:150]: yeah, above seems best ... I want backtraces for debugging, but maybe that can be a later feature

File: lib/kennel.rb [hunk 8]
```diff
@@ -100,26 +101,30 @@ def parts_serializer
 
     def generated
       @generated ||= begin
-        Progress.progress "Generating" do
+        parts = Progress.progress "Finding applicable projects and parts" do
```
grosser, line 104: says about the same ?
```suggestion
        parts = Progress.progress "Finding parts" do
```

lib/kennel.rb
```diff
@@ -100,26 +101,30 @@ def parts_serializer
...
           projects = projects_provider.projects
           projects = filter.filter_projects projects
 
           parts = Utils.parallel(projects, &:validated_parts).flatten(1)
-          parts = filter.filter_parts parts
+          filter.filter_parts parts
```
grosser, line 109: don't we need to re-assign here since filter does not modify ?

grosser (approved):
makes sense to me ... the safe_tracking_id seems a bit overkill but it's cheap so 🤷

lib/kennel.rb
zdrve: Sure, if you like

lib/kennel.rb
zdrve: That's done by `Progress.progress` returning the value of the block

PR #214, opened 2022-10-22
Limit number of lines of diff output per item

zdrve:
Dashboard widget diffs can be uselessly large.

image

### DIFF
--- a/lib/kennel/output_limiter.rb
+++ b/lib/kennel/output_limiter.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Kennel
+  class OutputLimiter
+    def initialize(out, max_lines, &on_overflow)
+      @out = out
+      @max_lines = max_lines
+      @on_overflow = on_overflow
+      @seen = 0
+    end
+
+    def puts(*content)
+      content.join("\n").split("\n").each do |line|
+        if @seen >= @max_lines
+          @on_overflow&.call
+          @on_overflow = nil
+          break
+        else
+          @out.puts(line)
+          @seen += 1
+        end
+      end
+    end
+  end
+end
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "diff/lcs"
+require_relative "./output_limiter"
 
 module Kennel
   class Syncer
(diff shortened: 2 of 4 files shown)


lib/kennel/output_limiter.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Kennel
+  class OutputLimiter
+    def initialize(out, max_lines, &on_overflow)
+      @out = out
+      @max_lines = max_lines
+      @on_overflow = on_overflow
```
grosser, line 8: simpler when overflow is a string and then get `puts`-ed when overflow happens ?

lib/kennel/syncer.rb
```diff
@@ -221,6 +221,14 @@ def print_plan(step, list, color)
     end
 
     def print_diff(diff)
+      max_diff_lines = [Integer(ENV.fetch("MAX_DIFF_LINES", "50")), 10].max
+      require_relative "./output_limiter"
```
grosser, line 225: prefer to require at the top, also avoid doing this multiple times

lib/kennel/output_limiter.rb
```diff
@@ -0,0 +1,24 @@
...
+      @seen = 0
+    end
+
+    def puts(*content)
+      content.join("\n").split("\n").each do |line|
+        if @seen >= @max_lines
+          @on_overflow&.call
+          @on_overflow = nil
```
grosser, line 16: can break here instead ?

grosser (approved):
I like the feature ... having a separate class+block feels a bit overkill, but that can be code-golfed later

lib/kennel/syncer.rb
```diff
@@ -221,6 +221,14 @@ def print_plan(step, list, color)
...
+      out = OutputLimiter.new(Kennel.out, max_diff_lines) do
+        message = "(Diff for this item truncated after #{max_diff_lines} lines. " \
+          "Rerun with MAX_DIFF_LINES=#{max_diff_lines * 2} to see more)"
+        Kennel.out.puts "  " + Utils.color(:magenta, message)
+      end
+
       diff.each do |type, field, old, new|
```
grosser, line 232: maybe we build a string here instead
... and then have a `truncate text, to: 100, with: "text was truncated blah blah"` method ?

lib/kennel/output_limiter.rb
zdrve: I'm not sure I see how it's simpler?

- it would lose the "&"
- s/call/puts/ (almost)

That's it, isn't it?

It would also be less flexible.

zdrve:
Re. overkill - syncer is too fat already. I'd like to reduce by how much we make it worse.

lib/kennel/syncer.rb
zdrve: Code Golf later :-)

PR #215, opened 2022-10-23
simplify output limiting

grosser:
code-golf for https://github.com/grosser/kennel/pull/214

- collect all the logic in a single (easy to read or ignore method)
- do not use complicated blocks/yields/facade pattern
- move feature tests into their own describe block
- remove the 10 line min ... let the users do what they want to do 🤷 
- use `split(/\n/)` to not remove "\n\n" cases

@zdrve


lib/kennel/syncer.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "diff/lcs"
-require_relative "./output_limiter"
```
zdrve, line 4: Arbitrarily adding my comments here because line-of-code comments are the only way to thread in GitHub.

I really want to make Syncer smaller, not bigger. It's already large and complex and hard to test; let's not make that worse. Why do you want to get rid of this class?

lib/kennel/syncer.rb
```diff
@@ -16,7 +15,6 @@ def initialize(api, expected, project_filter: nil, tracking_id_filter: nil)
       @project_filter = project_filter
       @tracking_id_filter = tracking_id_filter
       @expected = Set.new expected # need set to speed up deletion
-      @max_diff_lines = [Integer(ENV.fetch("MAX_DIFF_LINES", "50")), 10].max
```
zdrve, line 19: The 10 was there for a reason ... but feel free to reject the reasons, they are a bit silly.

- having a negative limit makes no sense, so that shouldn't be allowed
- having a zero limit makes sense, but then breaks the "Use N * 2 to see more" message
- having a limit of 1 makes sense and workse with "N * 2", but then causes the message to mention "1 lines", which is bad grammar

A minimum of 2 sidesteps all of those issues. Raising it to 10 was an arbitrary choice, primarily to distract attention from 2 being an apparently strange choice :-p

lib/kennel/syncer.rb
```diff
@@ -244,16 +234,19 @@ def print_diff(diff)
...
+              diff(old, new).map { |l| "    #{l}" }.join("\n")
+          elsif (old + new).size > 100
+            "  #{type}#{field}\n" \
+              "    #{old} ->\n" \
+              "    #{new}"
+          else
+            "  #{type}#{field} #{old} -> #{new}"
+          end
+
+        Kennel.out.puts truncate_diff(message)
```
zdrve, line 249: I like this diff – apart from the fact that I would prefer truncate_diff to be in a different file.

lib/kennel/utils.rb
```diff
@@ -57,6 +57,11 @@ def color(color, text, force: false)
         "\e[#{COLORS.fetch(color)}m#{text}\e[0m"
       end
 
+      def truncate_lines(text, to:, warning:)
+        return text if text.count("\n") < to
+        text.split(/\n/).first(to).join("\n") + "\n" + warning
```
zdrve, line 62: More efficient at runtime, harder to read:

```suggestion
        text.split(/\n/, to + 1).first(to).join("\n") + "\n" + warning
```

More than happy to keep it as it is though.

test/kennel/utils_test.rb
```diff
@@ -70,6 +70,24 @@
...
+
+    it "leaves short alone" do
+      call("a\nb").must_equal "a\nb"
+    end
+
+    it "truncates long" do
+      call("a\nb\nc").must_equal "a\nb\nSNIP!"
+    end
+
+    it "keeps sequential newlines" do
```
zdrve, line 86: Why is this desirable?

lib/kennel/utils.rb
zdrve: (Untested) Simpler, and also more consistent with newline handling maybe? Effectively requires that `text` and `warning` both end with a newline.

```suggestion
        lines = text.lines.take(to + 1)
        lines[-1] = warning if lines.length > to
        lines.join
```

lib/kennel/output_limiter.rb
```diff
@@ -1,25 +0,0 @@
...
-  class OutputLimiter
-    def initialize(out, max_lines, &on_overflow)
-      @out = out
-      @max_lines = max_lines
-      @on_overflow = on_overflow
-      @seen = 0
-    end
-
-    def puts(*content)
-      content.join("\n").split("\n").each do |line|
```
zdrve, line 13: > use split(/\n/) to not remove "\n\n" cases

I'm not sure that the old code used to do that, did it?

grosser: it did not
I did not see it causing an actual issue, just wanted the code to be technically correct

lib/kennel/utils.rb
grosser: I like just having 1 path through the code 👍

lib/kennel/syncer.rb
grosser: utils was meant for these 1-off methods that don't belong anywhere
once they get to complex or there are multiple that could be a class I try to move them out

lib/kennel/syncer.rb
grosser: kk, let's do 2 then :)

lib/kennel/syncer.rb
grosser: removing this class did not make syncer bigger
it still has the same logic as before "set max" + "build warning" + "call truncate"
(and now they are all in a single place, the `truncate_diff` method

zdrve: I'm not sure how it didn't make Syncer bigger. Syncer with the limiting built-in is more complex than Syncer without the limiting built in. Right?

lib/kennel/syncer.rb
zdrve: Suuuuure....  not sure why you mention that. It feels like you've answered a question that I didn't ask. Maybe I'm just tired.

lib/kennel/output_limiter.rb
zdrve: OK, just an odd thing to include in the PR description then.

lib/kennel/syncer.rb
zdrve: No wait I get it now, I misread something. Sorry.

zdrve (approved):

lib/kennel/syncer.rb
grosser: ah you meant truncate_diff not the utils method ...
the code was in syncer before so 🤷 
also having a new file for 5 loc makes reading the code harder
... as long as methods are self-contained and easy to read I don't mind having a lot of them, it just gets ugly if they all interact with each-other and share state

lib/kennel/output_limiter.rb
grosser: wanted to call out the diff

lib/kennel/syncer.rb
zdrve: Some people would call things that are self contained and don't share state with other things, a "class" :-p

grosser:
By that logic we should turn utils into 15 classes ... Going to be messy 🤗

PR #216, opened 2022-10-23
pre-download sketch

zdrve:
If it looks like we're going to get as far as syncer "plan" (and possibly "update"), then we're going to need to `download_definitions` from datadog, which can take quite a few seconds. So start it earlier, so that it's got a better chance of being ready in time.

Very small data sample of test runs on kennel-config master with no changes, running "rake generate plan":

- without this PR: approx 31-32 seconds
- with this PR: aprox 24-29 seconds

---

Nowhere near ready yet, and although it achieves the goal of making kennel faster, I've also broken updates (cache_metadata is used by syncer after creates). So there's a bit more work to do, plus testing etc.

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -14,6 +14,7 @@ require "kennel/projects_provider"
 require "kennel/syncer"
 require "kennel/id_map"
 require "kennel/api"
+require "kennel/downloader"
 require "kennel/github_reporter"
 require "kennel/subclass_tracking"
 require "kennel/settings_as_methods"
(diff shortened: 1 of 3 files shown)


grosser (commented):
the idea of starting the download early makes sense
I was expecting something like
```
def plan
  downloader.prepload
  ... rest of the code
```
and then accessing definitions in the dowloader waits until preload is complete and preload itself prints the "took xyz s", it should also print the "Started download" most likely

zdrve:
> I was expecting something like

It's not done that way because of `rake generate plan`, which I suspect is a common pattern. It's certainly my usual invocation.

PR #217, opened 2022-10-27
Use the correct kennel instance when checking strict_imports

zdrve:
The "Compatibility" mechanism makes things confusing, because it means that there is both a right and a wrong way to do things, instead of only a right way.

### DIFF
--- a/test/kennel/syncer_test.rb
+++ b/test/kennel/syncer_test.rb
@@ -105,9 +105,13 @@ describe Kennel::Syncer do
   let(:expected) { [] }
   let(:project_filter) { [] }
   let(:tracking_id_filter) { [] }
+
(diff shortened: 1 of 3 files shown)


test/kennel/syncer_test.rb
```diff
@@ -105,9 +105,13 @@ def add_identical
   let(:expected) { [] }
   let(:project_filter) { [] }
   let(:tracking_id_filter) { [] }
+
+  let(:kennel) { Kennel::Engine.new }
+
```
grosser, line 110: nothing else has the newlines 🤷 
```suggestion
  let(:kennel) { Kennel::Engine.new }
```

grosser (approved):

PR #218, opened 2022-10-27
Remove 'Compatibility'

zdrve:
Builds on #217 

Provide only the right way of doing things, not both a right way and a wrong way.

`out` and `err` remain in `Kennel` rather than in instances of `Kennel::Engine`, because moving them to the engine is a very wide-ranging change for rather little value.


lib/kennel.rb
```diff
@@ -84,11 +85,11 @@ def filter
     end
 
     def syncer
-      @syncer ||= Syncer.new(api, generated, project_filter: filter.project_filter, tracking_id_filter: filter.tracking_id_filter)
+      @syncer ||= Syncer.new(api, generated, kennel: self, project_filter: filter.project_filter, tracking_id_filter: filter.tracking_id_filter)
     end
 
     def api
-      @api ||= Api.new(ENV.fetch("DATADOG_APP_KEY"), ENV.fetch("DATADOG_API_KEY"))
```
grosser, line 91: nit: I kinda liked "Env stuff / global setup happens in kennel.rb", but not sure if that is still true

lib/kennel/tasks.rb
```diff
@@ -8,6 +8,10 @@
 module Kennel
   module Tasks
     class << self
+      def kennel
+        @kennel ||= Kennel::Engine.new
+      end
```
grosser, line 13: would it make more sense to have a `Kennel.engine` ?
seems a little strange that we have a global instance in the tasks ...

lib/kennel/tasks.rb
```diff
@@ -66,7 +70,7 @@ def git_push?
   # https://help.datadoghq.com/hc/en-us/requests/254114 for automatic validation
   desc "Verify that all used monitor  mentions are valid"
   task validate_mentions: :environment do
-    known = Kennel.send(:api)
+    known = Kennel::Api.new
```
grosser, line 73: `Kennel::Tasks.kennel.api` ?
... seems a little inefficient that we create many different instances of the api

test/kennel_test.rb
```diff
@@ -15,6 +15,8 @@ def write(file, content)
 
   let(:models_count) { 4 }
 
+  let(:kennel) { Kennel::Engine.new }
+
```
grosser, line 19:
```suggestion
  let(:kennel) { Kennel::Engine.new }

```

test/kennel/syncer_test.rb
```diff
@@ -105,9 +105,13 @@ def add_identical
   let(:expected) { [] }
   let(:project_filter) { [] }
   let(:tracking_id_filter) { [] }
+
+  let(:kennel) { Kennel::Engine.new }
+
```
grosser, line 110:
```suggestion
  let(:kennel) { Kennel::Engine.new }
```

grosser (commented):
seems like a step forward
still feels a bit inconsistent, but future PRs can fix that
not reusing the api seems like it could result in cache misses though

lib/kennel.rb
zdrve: ENV is read all over the place though. If there was a pattern that ENV is only read from kennel.rb then it's passed me by.

lib/kennel/tasks.rb
zdrve: I don't see why it would make more sense tbh. I'd prefer to have the library be more "clean", and the "default instance" (which is a form of state-sharing between rake tasks) handled by the rake tasks.

lib/kennel/tasks.rb
zdrve: I think the key word is "little". Show me that it makes a statistically meaningful difference and we discuss whether or not the breaking of encapsulation is worth it.

grosser: prefer to not introduce bugs on purpose
"show me that it makes **no** difference" 🤷

lib/kennel.rb
grosser: yeah I think over time this changed ... simpler to have it in api 👍

lib/kennel/tasks.rb
zdrve: Calling a private method using `send` is a bug.

grosser: it's not a "bug", but maybe bad api design :D
... can make it public 🤷

zdrve:
Ah, re. caches, now that might be a better argument. Maybe.

The only caching I can see is the FileCache stuff. All instances of Api would share the same cache. Is that a good thing? Well, yes, but it gets interesting if multiple APIs try to use it concurrently. It's already not thread-safe, I think, and I don't see any usage pattern which would allow interleaved single-thread access.

So I don't see a problem in terms of caches, neither from a corruption nor an efficient point of view.

lib/kennel/tasks.rb
zdrve: Sure, we can.

Please could either approve or state that that is the change you're asking for? I'm having trouble finding this actionable.

grosser (approved):
should be fine then if nobody needs these shared caches 🤞 
... I'd no be surprised if we have to walk this back in the future because we want some shared caches

lib/kennel/tasks.rb
grosser: it's not a change I would have done, so wanted to check that there is a good reason for it
since you checked the call sites to make sure the caching is not broken we should be good to go

grosser:
ready for release or more coming ?

PR #219, opened 2022-10-29
remove unnecessary wrapper classes

grosser:
- we don't care for what the plan was when doing an update (since the update should match the plan)
- we don't need to wrap update results since there is just 1 result (the actions performed)

@zdrve

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -69,12 +69,8 @@ module Kennel
     end
 
     def update
-      the_plan = syncer.plan
-      the_update = syncer.update if syncer.confirm
-      UpdateResult.new(
-        plan: the_plan,
-        update: the_update
-      )
+      syncer.plan
+      syncer.update if syncer.confirm
     end
 
     private
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -7,9 +7,6 @@ module Kennel
     DELETE_ORDER = ["dashboard", "slo", "monitor", "synthetics/tests"].freeze # dashboards references monitors + slos, slos reference monitors
     LINE_UP = "\e[1A\033[K" # go up and clear
 
-    Plan = Struct.new(:noop?, :no_change, :create, :update, :delete, keyword_init: true)
-    Update = Struct.new(:update_log, keyword_init: true)
-
     def initialize(api, expected, kennel:, project_filter: nil, tracking_id_filter: nil)
       @api = api
       @kennel = kennel
@@ -32,13 +29,9 @@ module Kennel
         print_plan "Delete", @delete, :red
       end
 
-      Plan.new(
-        noop?: noop?,
-        no_change: @no_change,
-        create: @create,
-        update: @update,
-        delete: @delete
-      )
+      @create.map { |_id, e| [:create, e.class.api_resource, e.tracking_id, nil] } +
+        @update.map { |_id, e| [:create, e.class.api_resource, e.tracking_id, nil] } +
+        @delete.map { |_id, _e, a| [:delete, a.fetch(:klass).api_resource, a.fetch(:tracking_id), a.fetch(:id)] }
     end
 
     def confirm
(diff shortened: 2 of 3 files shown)


lib/kennel/syncer.rb
```diff
@@ -8,7 +8,6 @@ class Syncer
     LINE_UP = "\e[1A\033[K" # go up and clear
 
     Plan = Struct.new(:noop?, :no_change, :create, :update, :delete, keyword_init: true)
-    Update = Struct.new(:update_log, keyword_init: true)
```
zdrve, line 11: Would prefer to keep this one since it allows for some extensibility. Returning _just_ an array of (rather odd nameless) structures makes it hard to add anything in future.

lib/kennel.rb
```diff
@@ -70,12 +69,8 @@ def plan
...
 
     def update
-      the_plan = syncer.plan
-      the_update = syncer.update if syncer.confirm
-      UpdateResult.new(
-        plan: the_plan,
-        update: the_update
-      )
+      syncer.plan
+      syncer.update if syncer.confirm
```
zdrve, line 73: `ci_and_notify` will need updating of course.

lib/kennel/syncer.rb
grosser: prefer to keep the api simple until we need anything else
also "what was updated" seems like the obvious return value for an "update method"
adding an extra struct does not make that simpler
```
actions = syner.update.actions
```
it not much better 🤷

grosser:
👁️ updated to make plan and update return the same action log so ci_and_notify will no longer have to have any internal kennel knowledge

grosser:
it could be nice to have a wrapper class for `Action` since that is just an array atm ... but `{action: :delete, foo: :bar}` would be good enough too ...

lib/kennel/syncer.rb
zdrve: I'm not arguing for keeping the "plan and update" level of wrapper. Feel free to get rid of that.

I am arguing for keeping the "actions" / "log" wrapper. i.e. don't just return an array, return something which responds to e.g. `:actions` to get that array.

grosser:
updated to return a `Plan` with `changes` from both ... sounds good ?
(was thinking about just returning an array of tracking-ids but that seemed like a bit to basic)

lib/kennel/syncer.rb
```diff
@@ -29,9 +30,12 @@ def plan
         print_plan "Delete", @delete, :red
       end
 
-      @create.map { |_id, e| [:create, e.class.api_resource, e.tracking_id, nil] } +
-        @update.map { |_id, e| [:create, e.class.api_resource, e.tracking_id, nil] } +
-        @delete.map { |_id, _e, a| [:delete, a.fetch(:klass).api_resource, a.fetch(:tracking_id), a.fetch(:id)] }
+      Plan.new(
+        changes:
+          @create.map { |_id, e| [:create, e.class.api_resource, e.tracking_id, nil] } +
+          @update.map { |_id, e| [:create, e.class.api_resource, e.tracking_id, nil] } +
```
zdrve, line 36:
```suggestion
          @update.map { |_id, e| [:update, e.class.api_resource, e.tracking_id, nil] } +
```

zdrve: Though it does suggest that maybe we should add test coverage for this

lib/kennel/syncer.rb
```diff
@@ -67,11 +71,11 @@ def update
         message = "#{klass.api_resource} #{a.fetch(:tracking_id)} #{id}"
         Kennel.out.puts "Deleting #{message}"
         @api.delete klass.api_resource, id
-        actions << [:delete, klass.api_resource, a.fetch(:tracking_id), id]
+        changes << [:delete, klass.api_resource, a.fetch(:tracking_id), id]
         Kennel.out.puts "#{LINE_UP}Deleted #{message}"
       end
 
-      actions
+      Plan.new(changes: changes)
```
zdrve, line 78: 👍

zdrve (approved):
Consider adding tests for the returned Plan (but hey I started the bad habit ...)

👍

grosser:
is this ok to release or do we need more validation PRs ?

PR #222, opened 2022-10-30
Fail / Report modes for `validate: false`

zdrve:
Working towards error tagging, i.e. more fine-grained error suppression.

In terms of functionality, what this PR adds is new behaviours for `validate: false`: 

If `validate: false` is used, but there are no validation errors (i.e. you could set `validate: true` and it would still work), then what happens is governed by env var `UNNECESSARY_VALIDATE_FALSE`:

 * `fail` (causes a validation error)
 * `show` (print to stdout and continue)
 * (everything else) ignore

The default is `fail` if filtering is in use (but maybe `show` would be better?), otherwise ignore.

(_Update_: I removed all of the unnecessary `validate: false`s the other day, so the above is not going to hit much any more)

If `validate: false` is used and there _are_ validation errors, then what happens is governed by env var `SUPPRESSED_ERRORS`:

 * `fail` (causes validation to fail anyway, i.e. effectively sets `validate` back to true)
 * `show` (print to stdout and continue)
 * (everything else) ignore

The default is `ignore`.

### DIFF
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -156,7 +156,7 @@ module Kennel
       def as_json
         # A courtesy to those tests that still expect as_json to perform validation and raise on error
         build if @validation_errors.nil?
-        raise Kennel::ValidationError, "#{safe_tracking_id} #{@validation_errors.first}" unless validation_errors.empty?
+        raise Kennel::ValidationError, "#{safe_tracking_id} #{@validation_errors.first.message}" unless validation_errors.empty?
 
         @as_json
       end
@@ -209,7 +209,7 @@ module Kennel
       end
 
       def invalid!(message)
-        validation_errors << message
+        validation_errors << ValidationMessage.new(message)
       end
 
       def raise_with_location(error, message)
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -150,13 +150,7 @@ module Kennel
           end
         end
 
-        @filtered_validation_errors =
-          if validate
-            unfiltered_validation_errors
-          else
-            []
-          end
-
+        @filtered_validation_errors = filter_validation_errors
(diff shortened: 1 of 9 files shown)


zdrve:
Zendesk kennel has 372 "suppressed errors" and 126 "no validation errors to suppress".

zdrve:
Nudge, @grosser :-)

test/kennel/optional_validations_test.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 require_relative "../test_helper"
 
-SingleCov.covered!
+SingleCov.covered! uncovered: 19
```
zdrve, line 4: I got lazy. Most of this code goes away in the next PR. Let me know if that's not good enough.

lib/kennel/models/record.rb
```diff
@@ -136,27 +138,28 @@ def build_json
...
-        @as_json = json # Only valid if validation_errors.empty?
+        @filtered_validation_errors = filter_validation_errors
+        @as_json = json # Only valid if filtered_validation_errors.empty?
       end
 
       def as_json
         # A courtesy to those tests that still expect as_json to perform validation and raise on error
-        build if @validation_errors.nil?
-        raise Kennel::ValidationError, "#{safe_tracking_id} #{@validation_errors.first}" unless validation_errors.empty?
+        build if @unfiltered_validation_errors.nil?
```
grosser, line 161: this would make more sense ?
```suggestion
        build if @as_json.nil?
```

grosser:
thx for keeping it small :D

lib/kennel/models/record.rb
zdrve: I don't think it makes any difference, but I haven't checked carefully.

PR #224, opened 2022-11-15
Fix plan update tag

zdrve:
* Updates should be return as `:update` not `:create`
* Represent changes as structs (`c.type, c.id, ...`) instead of 4-element arrays

Backwards incompatible. Will require changes in Zendesk `ci_and_notify`

----

The failing test is failing on kennel-lib `master` :-(


grosser:
can reproduce the error so will make a separate PR

PR #228, opened 2022-11-28
support monitor variables

grosser:
log monitors can use `variables` when combining multiple queries so we have to store them

testing:
- does not influence other monitors
- does not influence log monitors
- works when updating log monitors

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -21,7 +21,8 @@ module Kennel
         renotify_interval: 0,
         notify_audit: false,
         no_data_timeframe: nil, # this works out ok since if notify_no_data is on, it would never be nil
-        groupby_simple_monitor: false
+        groupby_simple_monitor: false,
+        variables: nil
       }.freeze
       DEFAULT_ESCALATION_MESSAGE = ["", nil].freeze
       ALLOWED_PRIORITY_CLASSES = [NilClass, Integer].freeze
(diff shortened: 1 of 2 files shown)


zdrve:
To help review this, could you provide an example (or pointer to one) of variables in use?

grosser:
```
  variables: -> {
    [
      {
        data_source: "logs",
        name: "query",
        search: {
          query: "service:billing env:production @http.method:PUT"
        },
        indexes: [
          "*"
        ],
        compute: {
          aggregation: "count"
        },
        group_by: []
      },
      {
        data_source: "logs",
        name: "query1",
        search: {
          query: "service:billing env:production @http.method:POST"
        },
        indexes: [
          "*"
        ],
        compute: {
          aggregation: "count"
        },
        group_by: []
      }
    ]
  },
```

zdrve:
and what does that look like _in use_? I'm just completely failing to understand what variables are _for_ (having never seen them before and having failed to find any Datadog docs on them). And I'd like to understand what they look like, including _how_ they are used and _why_, to give proper context for reviewing this PR in any way more meaningful than "can confirm, is a setting".

grosser:
they are used in log monitor queries like so:
```
  query: -> { "formula(\"((query - query1) / query) * 100\").last(\"2h\") < #{critical}" },
```
basically a way of expressing sub-queries without building a long query string

zdrve:
OK, thanks! In which case:

- should we have a check to ensure that, as a minimum, the `name`s of variables are unique?
- what should the normalisation logic for variables look like? (are `[]` and `nil` and _no value_ equivalent? Does the order of the array matter?)
- should we have a check to ensure that there are no unused variables?

It might be OK to say "OK, maybe, in a later PR", but I just want to work out what we might want to add, and _why_ (to mitigate risks of CI failure due to lack of normalisation, maybe?), and therefore work out what needs doing now / what can be done later.

grosser:
we don't check names for dashboards either (which use a similar syntax) and it has not been an issue so far
... so all of these can be done later + "when needed"

grosser:
none of these concerns seem to be blockers, so merging to unblock these monitors for now

PR #229, opened 2022-11-30
update ruby to 3.1.2 to match mac ventura

grosser:
... and will give us a solid speed boost + prep us for the future

/cc @zdrve


lib/kennel/utils.rb
```diff
@@ -5,6 +5,7 @@ module Utils
 
     class TeeIO < IO
       def initialize(ios)
+        super(0)
```
zdrve, line 8: That's an unfamiliar construct and worth a code comment, I think. Why 0?

grosser: ah yeah ... a bogus value to make it stop complaining :D
... I'll add a comment

grosser: updated

lib/kennel/utils.rb
```diff
@@ -5,6 +5,7 @@ module Utils
 
     class TeeIO < IO
       def initialize(ios)
+        super(0) # called with fake file descriptor 0, so we can call super and get a proper class
```
zdrve, line 8: FD 0 isn't "fake", it's stdin. Does it _matter_ that it's stdin? Would 1 (stdout) work just as well? If not, why not? Does this even need to be a subclass of `IO`?

The code _might_ be fine; the comment leaves a lot of questions unanswered.

zdrve: I'm also of course aware that `TeeIO` isn't a new class, but the point stands regardless.

grosser:
It needs to inherit from Io to get all it's methods.
Not really the point of this pr, can be its own issue if you think it's
worth discussing more, it works since a while and has not made any trouble
so I would just leave it alone for now.

zdrve (approved):
I'm curious about the `IO` thing now but I'll ponder that outside of this PR :-) 

Anyway. I notice that currently, kennel-config and kennel-lib do not have matching `.ruby-version`s. Should we fix that? Is that idea of this PR that we update k-config to 3.1.2 also, so that all kennel users are running 3.1.2?

grosser:
yes, both should be in sync
still need to confirm the exact ventura ruby version and it's rollout schedule 🤞

PR #231, opened 2022-12-05
pre-download, again

zdrve:
If we're going to need to do a download, then trigger the download in the background so it can run while we're doing the generate.

### DIFF
--- a/test/kennel/downloader_test.rb
+++ b/test/kennel/downloader_test.rb
@@ -1,0 +1,46 @@
+# frozen_string_literal: true
+require_relative "../test_helper"
+
+SingleCov.covered!
+
+describe Kennel::Downloader do
+  let(:api) { stub("api") }
+  let(:downloader) { Kennel::Downloader.new(api) }
+
+  let(:dashboards) { [item, item] }
+  let(:monitors) { [item, item] }
+  let(:slos) { [item, item] }
+  let(:synthetics) { [item, item] }
+
+  before do
+    api.stubs(:list).once.with("dashboard", anything).returns(dashboards: dashboards)
+    api.stubs(:list).once.with("monitor", anything).returns(monitors)
+    api.stubs(:list).once.with("slo", anything).returns(data: slos)
+    api.stubs(:list).once.with("synthetics/tests", anything).returns(synthetics)
+  end
+
+  def item
+    # Not realistic IDs (wrong type and/or wrong format),
+    # but that shouldn't matter for these tests
+    @seq = (@seq || 0) + 1
+    { id: @seq }
+  end
+
+  it "downloads" do
+    answer = downloader.all_by_class
+    expected = {
+      Kennel::Models::Dashboard => dashboards,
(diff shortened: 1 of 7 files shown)


test/kennel/downloader_test.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+require_relative "../test_helper"
+
+SingleCov.covered! uncovered: 9
+
+describe Kennel::Downloader do
```
zdrve, line 6: TODO: write at least one test.

grosser:
still prefer https://github.com/grosser/kennel/pull/221 but need to see if the bug is fixable 😞

zdrve:
At this point I'd really like to hear a much stronger argument for preferring #221, because the only one I think I've heard so far is "it's simpler", and ... well, it just isn't. Not by any metric that I can see.

grosser:
it adds a `parallel` around 2 existing method calls and moves 1 method
vs reworking the control flow, adding a new class, integrating that class into 3 places
also it's "+30 loc" with tests

basically the only actual logic change is
```
        expected = generated
        actual = download_definitions
```
->
```
        expected, actual = Utils.parallel([:generated, :download_definitions]) { |m| send m }

```
the rest is just moving logic around and not adding new conditionals or state that needs to be checked

first commit is [moving 1 method around](https://github.com/grosser/kennel/commit/7af542bb5758050956313bcec6cdb0b92f157539)
2nd commit [adds parallel](https://github.com/grosser/kennel/commit/5a9ee8af768938bd46b1709cb53d48741fb6a8df)

zdrve:
Your PR touches roughly twice as many files, and twice as many lines of code, and has taken two PRs and 1 revert so far.

Please explain in what way my PR is worse.

grosser:
- it took 2 PRs because I added docs+made the api nicer, so not important
- it had a bug so I reverted it, might indicate that it will be tricky to get right
- it does not make the core class know about rake tasks
- it does not need a mutex
- it reuses the progress logic instead of duplicating it
- it adds far less code
- it touches 1 more file (it touches progress+test because I want to reuse it's logic, so not counting that)
- it does not introduce ambiguous state (downloader) but terminates the parallelism right there in the syncer
- it adds 1 cyclomatic complexity (via https://github.com/soutaro/cycromatic + https://github.com/grosser/pru `cycromatic lib | pru 'Integer(self[/\d+$/])' sum`) vs `11`

zdrve:
Fine, have it your way. This PR is ready to go.

grosser:
I presented what I think is a good list of arguments and took time looking into both approaches.
I understand if you don't want to spend more time on a "either way works 🤷 " situation, but I hope you see that I'm not just picking "option X because I wrote it".
I looked at the initial PR and though "hmm that's neat, but maybe it can be even better" and found a way that I think archives that.

grosser:
refactored it now make the "preloading" more explicit and fix the storage bug

grosser:
hope that's the last we see from this issue 🤞

PR #234, opened 2022-12-10
Builder vs Built

zdrve:
This PR touches a lot of files, but there's hardly any new code.

Splits "Record" and its subclasses into "Builder" and "Built" halves.

Motivation for this PR:

The lifetime of a "Record" instance (i.e. a monitor, dashboard, etc) is very roughly:

- instantiate with settings
- call "build", which runs build_json and validate_json
- potentially store in `./generated`, and run in the Syncer for a plan / update

If one analyzes which methods are called pre- and post- "build" (or more specifically, pre- and post- the moment that build finishes), then we see a clear pattern:

- These methods are _only_ called post "build": resolve_linked_tracking_ids!, add_tracking_id, filtered_validation_errors, tracking_id?, resolve, as_json, diff, resolve_link
- A few methods are called both pre- and post-build: project, tracking_id, id
- Everything else is _only_ called pre-build

This suggests that the pre- and post-build objects should actually be in separate classes; and that's what this PR implements. Splitting a big class into two smaller parts with a clear integration between the two, makes both parts easier to work with.

The junction between the two halves is here:
https://github.com/grosser/kennel/blob/zdrve/20221209-built/lib/kennel/models/record.rb#L134-L150 calls https://github.com/grosser/kennel/blob/zdrve/20221209-built/lib/kennel/models/built/record.rb#L7-L21 .

Notable mostly-unrelated change also in this PR: [with_preserved_subclass_tracking](https://github.com/grosser/kennel/blob/zdrve/20221209-built/test/test_helper.rb#L46-L62) , which allows tests to instantiate temporary test subclasses without messing up the `subclasses` output for later tests.

Things I'm not wild about in this PR, and would welcome opinions on even more than the rest of the PR: 
- the name "unbuilt_class"
- the fact that we now have two classes called Record, two called Monitor, and so on. The Built classes could just as well be called "BuiltRecord", "BuiltMonitor" and so on. Maybe that would be better.

### DIFF
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -30,6 +30,12 @@ require "kennel/models/monitor"
 require "kennel/models/slo"
 require "kennel/models/synthetic_test"
 
+require 'kennel/models/built/record'
+require 'kennel/models/built/dashboard'
+require 'kennel/models/built/monitor'
+require 'kennel/models/built/slo'
+require 'kennel/models/built/synthetic_test'
+
 # settings
 require "kennel/models/project"
 require "kennel/models/team"
@@ -121,7 +127,7 @@ module Kennel
           ERROR
         end
 
-        Progress.progress "Building json" do
+        parts = Progress.progress "Building json" do
           # trigger json caching here so it counts into generating
           Utils.parallel(parts, &:build)
         end
--- a/lib/kennel.rb
+++ b/lib/kennel.rb
@@ -30,11 +30,11 @@ require "kennel/models/monitor"
 require "kennel/models/slo"
 require "kennel/models/synthetic_test"
 
-require 'kennel/models/built/record'
-require 'kennel/models/built/dashboard'
-require 'kennel/models/built/monitor'
-require 'kennel/models/built/slo'
-require 'kennel/models/built/synthetic_test'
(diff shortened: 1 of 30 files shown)


zdrve:
Leaving as draft for now because I want to run this codebase locally for a day or two first.

grosser (commented):
I prefer "monitor logic" to be in a single "Monitor" class.
With this change I'd have to look into 2 places and make these 2 places copy-paste logic if they ever overlap or need some shared logic.
Also every contributor needs to understand 1 more concept when trying to add something whereas before they could just go to monitor.rb and cmd+f for what they were looking for.

grosser:
alternative could be grouping these methods together or having some kind of
```
# methods that are used on built objects
```
possibly a shared `Monitor` with some `include Monitor::BuilderMethods` (like we do in rails)
none of these classes feel overwhelming to me atm, and they are not 2 different `Record`s, they are the same `Record` just different concerns

zdrve:
> built objects

That's ... that's what a class is.

We could put them in separate classes in the same file, would that help?

grosser:
Same issue, they'd still be separate records. Can validations be a module?
Then the simple ones could avoid this complexity.

PR #235, opened 2022-12-11
refactor syncer logic more

grosser:
mostly more comments but also logically ordering things more and not relying on modifying the instance vars

/cc @zdrve

### DIFF
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -122,53 +122,62 @@ module Kennel
 
     def calculate_changes
       @warnings = []
-      @update = []
-      @delete = []
       @id_map = IdMap.new
 
       Progress.progress "Diffing" do
         populate_id_map @expected, @actual
         filter_actual! @actual
         resolve_linked_tracking_ids! @expected # resolve dependencies to avoid diff
+        @expected.each(&:add_tracking_id) # avoid diff with actual, which has tracking_id
 
-        @expected.each(&:add_tracking_id) # avoid diff with actual
+        # see which expected match the actual
+        matching, unmatched_expected, unmatched_actual = partition_matched_expected
+        validate_expected_id_not_missing unmatched_expected
+        fill_details! matching # need details to diff later
 
-        lookup_map = matching_expected_lookup_map
-        items = @actual.map do |a|
-          e = matching_expected(a, lookup_map)
-          if e && @expected.delete?(e)
-            [e, a]
-          else
-            [nil, a]
-          end
-        end
-
-        # fill details of things we need to compare
-        details = items.map { |e, a| a if e && e.class.api_resource == "dashboard" }.compact
-        @api.fill_details! "dashboard", details
-
-        # pick out things to update or delete
-        items.each do |e, a|
+        # update matching if needed
+        @update = matching.map do |e, a|
           id = a.fetch(:id)
-          if e
-            diff = e.diff(a) # slow ...
-            if diff.any?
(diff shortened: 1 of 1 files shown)


lib/kennel/syncer.rb
```diff
@@ -130,45 +130,56 @@ def calculate_changes
...
-            [e, a]
-          else
-            [nil, a]
-          end
+        # update matching if needed
+        matching.each do |e, a|
+          id = a.fetch(:id)
+          diff = e.diff(a)
+          @update << [id, e, a, diff] if diff.any?
         end
```
zdrve, line 145:
```suggestion
        @update = matching.map do |e, a|
          id = a.fetch(:id)
          diff = e.diff(a)
          [id, e, a, diff] if diff.any?
        end.compact
```

Does that work?

lib/kennel/syncer.rb
```diff
@@ -130,45 +130,56 @@ def calculate_changes
...
 
-        # fill details of things we need to compare
-        details = items.map { |e, a| a if e && e.class.api_resource == "dashboard" }.compact
-        @api.fill_details! "dashboard", details
+        # delete previously managed
+        unmatched_actual.each { |a| @delete << [a.fetch(:id), nil, a] if a.fetch(:tracking_id) }
```
zdrve, line 148: Likewise,

```suggestion
        @delete = unmatched_actual.map { |a| [a.fetch(:id), nil, a] if a.fetch(:tracking_id) }.compact
```

lib/kennel/syncer.rb
```diff
@@ -130,45 +130,56 @@ def calculate_changes
...
+      matched = []
+      @actual.each do |a|
+        e = matching_expected(a, lookup_map)
+        if e && unmatched_expected.delete?(e)
+          matched << [e, a]
+        else
+          unmatched_actual << a
+        end
+      end.compact
+      [matched, unmatched_expected, unmatched_actual]
```
zdrve, line 172: I like this function :-) 

Mild preference for returning a struct so one can't accidentally get the ordering of the triple wrong.

e.g. `Struct.new(:matched, :unmatched_expected, :unmatched_actual).new(matched, unmatched_expected, unmatched_actual`

It's more verbose, but safer.

lib/kennel/syncer.rb
```diff
@@ -130,45 +130,56 @@ def calculate_changes
...
+    end
+
+    # fill details of things we need to compare
+    def fill_details!(details_needed)
+      details_needed = details_needed.map { |e, a| a if e && e.class.api_resource == "dashboard" }.compact
+      @api.fill_details! "dashboard", details_needed
+    end
+
+    def validate_expected_id_not_missing(expected)
```
zdrve, line 181: A better name, I approve 👍🏼

zdrve (approved):

lib/kennel/syncer.rb
grosser: ah yes ✨ 
`filter_map` ideally once we are past 2.6

lib/kennel/syncer.rb
grosser: just looks so ugly 😭

lib/kennel/syncer.rb
grosser: yeah much better now

lib/kennel/syncer.rb
grosser: all the code is "expected, actual" so doubt the ordering will go wrong 🤞

PR #236, opened 2022-12-12
speed up diffing by using builtin ==

grosser:
before:
```
Diffing ... 10.4s
```

after:
```
Diffing ... 2.86s
```

/cc @zdrve


zdrve (approved):
I've often wondered why the code didn't already do this!

PR #238, opened 2022-12-12
warn when updating

grosser:
somehow we only warned when planning ... but should always do it

@zdrve


grosser:
yeah good point, will address in it's own PR

PR #240, opened 2022-12-14
make PartsSerializer more symmetric

grosser:
little bit of random cleanup and then making PartsSerializer use the same old_paths method for all cases

### DIFF
--- a/lib/kennel/parts_serializer.rb
+++ b/lib/kennel/parts_serializer.rb
@@ -8,13 +8,9 @@ module Kennel
 
     def write(parts)
       Progress.progress "Storing" do
-        if filter.tracking_id_filter
-          write_changed(parts)
-        else
-          old = old_paths
-          used = write_changed(parts)
-          (old - used).uniq.each { |p| FileUtils.rm_rf(p) }
-        end
+        existing = existing_paths
+        used = write_changed(parts)
+        FileUtils.rm_rf((existing - used).uniq)
       end
     end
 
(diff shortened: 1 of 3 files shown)


lib/kennel/parts_serializer.rb
```diff
@@ -26,31 +22,36 @@ def write_changed(parts)
...
     end
 
-    def old_paths
-      Dir["{#{directories_to_clean_up.join(",")}}/**/*"]
+    def path_for_tracking_id(tracking_id)
+      "generated/#{tracking_id.tr("/", ":").sub(":", "/")}.json"
     end
 
     def write_file_if_necessary(path, content)
+      # note: always generating in faster than JSON.load-ing and comparing
```
zdrve, line 52: s/in/is/ ?

lib/kennel/parts_serializer.rb
```diff
@@ -8,13 +8,9 @@ def initialize(filter:)
...
-        if filter.tracking_id_filter
-          write_changed(parts)
-        else
-          old = old_paths
-          used = write_changed(parts)
-          (old - used).uniq.each { |p| FileUtils.rm_rf(p) }
-        end
+        old = old_paths
+        used = write_changed(parts)
+        (old - used).uniq.each { |p| FileUtils.rm_rf(p) }
```
zdrve, line 13:
```suggestion
        File.unlink((old - used).uniq)
```

I think. It might be worth add a code comment / clarifying variable name to point out that `used` is a mixture of files and directories, whereas `old` is only files. In fact, come to think of it: we should make that consistent. I don't think that `used` needs to contain the directories, since we seem to rely on the fact that empty directories disappear via `git`.

grosser: - old is folders too `Dir["generated/**/*"].grep_v(/json/)` has loads
- git does not remove empty folders so it's nice if we do it

improved the names a bit `old -> existing` and using `FileUtils.rm_rf list`

grosser:
merging this for now ... more changes can be their own PR

lib/kennel/parts_serializer.rb
```diff
@@ -26,31 +22,36 @@ def write_changed(parts)
...
+        write_file_if_necessary(path, content)
       end
 
       used
     end
 
-    def directories_to_clean_up
-      if filter.project_filter
-        filter.project_filter.map { |project| "generated/#{project}" }
+    def existing_paths
```
zdrve, line 37: I'm not a fan of this change (which got merged already :-( ) because it's asymmetric and misleading. The first two branches only return files, not directories; the third returns both. Does that difference matter? It's not clear. The method name doesn't make that clear either.

grosser: ah I thought your only concern was refactoring the paths even more :(

grosser: it returns "files or folders" 🤷 
_paths is kinda vague for that reason
`_files_or_folders` is kinda ugly

lib/kennel/parts_serializer.rb
```diff
@@ -8,13 +8,9 @@ def initialize(filter:)
...
-        if filter.tracking_id_filter
-          write_changed(parts)
-        else
-          old = old_paths
-          used = write_changed(parts)
-          (old - used).uniq.each { |p| FileUtils.rm_rf(p) }
-        end
+        existing = existing_paths
+        used = write_changed(parts)
+        FileUtils.rm_rf((existing - used).uniq)
```
grosser, line 13: ... the .uniq should not be needed

zdrve: In what sense? That the data can never contain duplicates, or that it's OK to allow duplicates through?

lib/kennel/parts_serializer.rb
grosser: https://github.com/grosser/kennel/pull/241

lib/kennel/parts_serializer.rb
grosser: existing could never have duplicates since it's just a list of files on disk

zdrve: Right (but only because we `uniq` the project / tracking_id filters); and `used` can, but that doesn't matter because that's the removal list. Got it 👍

PR #241, opened 2022-12-15
files and folders

grosser:


zdrve (approved):
Yeah I think that's probably a good enough clue as to what's going on. It's a bit too subtle otherwise.

The last time we had a problem in this area we improved the test coverage, so this PR / yesterday's PR are probably functionally fine, this is just about maintainability.

PR #242, opened 2022-12-21
Split Utils

zdrve:
Splitting up the grab-bag "Utils" class into three:

 - string stuff
 - console / UI stuff
 - the rest

which turns out to make about three equal-sized parts.

Potentially a **BREAKING CHANGE** for anything using `Utils`.

### DIFF
--- a/lib/kennel/console_utils.rb
+++ b/lib/kennel/console_utils.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+module Kennel
+  module ConsoleUtils
+    COLORS = { red: 31, green: 32, yellow: 33, cyan: 36, magenta: 35, default: 0 }.freeze
+
+    class TeeIO < IO
+      def initialize(ios)
+        super(0) # called with fake file descriptor 0, so we can call super and get a proper class
+        @ios = ios
+      end
+
+      def write(string)
+        @ios.each { |io| io.write string }
+      end
+    end
+
+    class << self
+      def ask(question)
+        Kennel.err.printf color(:red, "#{question} -  press 'y' to continue: ", force: true)
+        begin
+          STDIN.gets.chomp == "y"
+        rescue Interrupt # do not show a backtrace if user decides to Ctrl+C here
+          Kennel.err.print "\n"
+          exit 1
+        end
+      end
+
+      def color(color, text, force: false)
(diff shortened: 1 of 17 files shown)


lib/kennel/string_utils.rb
```diff
@@ -0,0 +1,39 @@
...
+          .gsub(/^-|-$/, "") # remove leading/trailing
+      end
+
+      def truncate_lines(text, to:, warning:)
+        lines = text.split(/\n/, to + 1)
+        lines[-1] = warning if lines.size > to
+        lines.join("\n")
+      end
+
+      def natural_order(name)
```
grosser, line 34: idk about this one ... feels like the wrong place

grosser (commented):
the added complexity from now having to find where things are is kinda meh
... especially when it's not super-clear like `color` could be in console or string
and then over time things might move and more breaks :(

... I really liked the grouping of the _case/parameterize methods though
... maybe step 1 is split it here but extend them all into `Utils` so maintenance is simpler but the UI stays the same

lib/kennel/console_utils.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+module Kennel
+  module ConsoleUtils
```
grosser, line 3: maybe name this `Console` ?
`Kennel::Console.ask?` is kinda nice

lib/kennel/string_utils.rb
zdrve: Good point, the others are all string->string functions and this one is something very different.

zdrve:
> ... especially when it's not super-clear like color could be in console or string

True insofar as it's both a string transform, and to do with console stuff... the distinction I see being that StringUtils is more generally-useful stuff (e.g. transforming strings to go into Datadog monitor titles and messages and so on), whereas you would never use `color` in such a context.

> the added complexity from now having to find where things are is kinda meh
> and then over time things might move and more breaks :(

I don't really buy these arguments, if that's what they are. One could also argue that having a file / module which is three times as large and contains unrelated things is "added complexity", i.e. this PR _reduces_ complexity. Ultimately, it's a trade-off; I haven't, for example, split it into 15 or so modules with one method each ;-)

PR #243, opened 2022-12-22
Get rid of Utils.inline_resource_metadata

zdrve:
Have `Api` handle it internally.

i.e. every object coming back from `show`, `list`, `create`, or `update` responds to `#klass` (e.g. `Kennel::Models::Monitor`), and (somewhat redundantly) `#tracking_id`).

Slightly less miscellanous code in `Utils`, slightly more in `Api`. More symmetrical code in `Syncer`.

### DIFF
--- a/lib/kennel/api.rb
+++ b/lib/kennel/api.rb
@@ -16,7 +16,7 @@ module Kennel
       response = request :get, "/api/v1/#{api_resource}/#{id}", params: params
       response = response.fetch(:data) if api_resource == "slo"
       response[:id] = response.delete(:public_id) if api_resource == "synthetics/tests"
-      response
+      tag(api_resource, response)
     end
 
     def list(api_resource, params = {})
@@ -32,7 +32,7 @@ module Kennel
         # ignore monitor synthetics create and that inherit the kennel_id, we do not directly manage them
         response.reject! { |m| m[:type] == "synthetics alert" } if api_resource == "monitor"
 
-        response
+        response.map { |r| tag(api_resource, r) }
       end
     end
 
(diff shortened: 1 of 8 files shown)


lib/kennel/api.rb
```diff
@@ -145,5 +145,23 @@ def with_cache(enabled, key)
         result
       end
     end
+
+    def tag(api_resource, reply)
+      ResourceTaggedHash.new(api_resource, reply)
+    end
+
+    class ResourceTaggedHash < Hash
```
grosser, line 153: careful with subclassing hash it can get weird ... for example after merging methods might be gone etc
also check performance in case things get slower when not using native hashes

grosser (commented):
I like the idea, but can we put the symbols into the hash instead to keep things native ?

zdrve:
> I like the idea, but can we put the symbols into the hash instead to keep things native?

Which bit of the idea do you like? What do you mean by "to keep things native", and why is that desirable?

grosser:
I like the moving of the logic.
Keep native -> use a hash and add extra symbol keys.
Why it's good: avoids any kind of funkyness with `merge` etc hash methods returning a hash that is not of the same class and might be better for performance since ruby might have some internal hackery that makes native/C-hashes faster.
... also it's the smaller change and we can revisit wrapping in in a class later if it's still an improvement

lib/kennel/api.rb
```diff
@@ -5,6 +5,14 @@ module Kennel
   class Api
     CACHE_FILE = "tmp/cache/details"
 
+    def self.tag(api_resource, reply)
```
grosser, line 8: can this be an instance method since all do `self.class.` ?

grosser (approved):
🎉

lib/kennel/api.rb
zdrve: Yes, or it can be a class method. It doesn't use any instance data.

PR #244, opened 2023-01-05
Ensure API methods still work with unmapped resources

adammw:
#243 introduced incompatibility with api_resources not in the api_resource_map.

This restores the previous behaviour if the klass is unknown.

cc @grosser @zdrve

### DIFF
--- a/lib/kennel/api.rb
+++ b/lib/kennel/api.rb
@@ -7,10 +7,12 @@ module Kennel
 
     def self.tag(api_resource, reply)
       klass = Models::Record.api_resource_map[api_resource]
-      klass ? reply.merge(
+      return reply unless klass # do not blow up on unknown models
+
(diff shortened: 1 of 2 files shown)


lib/kennel/api.rb
```diff
@@ -7,10 +7,10 @@ class Api
 
     def self.tag(api_resource, reply)
       klass = Models::Record.api_resource_map[api_resource]
```
grosser, line 9:
```suggestion
      klass = Models::Record.api_resource_map[api_resource]
      return reply unless klass # do not blow up on unknown models
```

grosser (approved):

zdrve (approved):
+1 with @grosser's change / fix.

zdrve:
Thanks @adammw :-)

PR #249, opened 2023-01-08
Break up 'Syncer'

zdrve:
This PR:

- removes the always-unused first argument to `validate_update!`
- rearranges the code in Syncer, while leaving the tests unchanged

The rearrangement mostly takes the form of creating several child modules and classes under `Syncer`, containing the same code that used to be in `Syncer`. These new sub-modules/classes remain in `syncer.rb` for now, not least precisely so we can do this refactoring without having to change any tests.

 - `MatchedExpected`, which deals with matching `expected` against `actual`
 - `Resolver`, which does everything related to `IdMap` and resolving Kennel IDs to Datadog IDs
 - `PlanDisplayer`, which handles `print_plan`

Finally there's a `Types` module, and a series of classes (`PlannedCreate` etc) to replace the old `|id, e, a, diffs|` quad.

All of this means that `Syncer` itself (i.e. without the submodules/classes) is now down to only 180 lines (from 300), and easier to follow; and that the code that's there is more readable and symmetrical.

If we wish then in later PRs we might want to (1) refactor the testing such that Syncer is tested separately from the new mods/classes, and (2) move those mods/classes (and their new tests) to their own files.

Presented as individual commits for your reviewing pleasure. ;-)

### DIFF
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -69,7 +69,7 @@ module Kennel
         reply = @api.create e.class.api_resource, e.as_json
         id = reply.fetch(:id)
         changes << Change.new(:create, e.class.api_resource, e.tracking_id, id)
-        @resolver.populate_id_map [], [reply] # allow resolving ids we could previously no resolve
+        @resolver.add_actual [reply] # allow resolving ids we could previously no resolve
         Kennel.out.puts "#{LINE_UP}Created #{message} #{e.class.url(id)}"
       end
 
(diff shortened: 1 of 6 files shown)


lib/kennel/syncer.rb
```diff
@@ -147,18 +144,19 @@ def validate_expected_id_not_missing(expected)
 
     def print_changes(step, list, color)
       return if list.empty?
-      list.each do |_, e, a, diff|
-        klass = (e ? e.class : a.fetch(:klass))
-        Kennel.out.puts Console.color(color, "#{step} #{klass.api_resource} #{e&.tracking_id || a.fetch(:tracking_id)}")
-        diff&.each { |args| Kennel.out.puts @attribute_differ.format(*args) } # only for update
+      list.each do |item|
+        Kennel.out.puts Console.color(color, "#{step} #{item.api_resource} #{item.tracking_id}")
+        if item.respond_to?(:diff)
```
grosser: can we have nil diff instead and do `item.diff&.each` ?

grosser (approved):
all looks good
the only kinda red flag for me was the ~100 LOC added by splitting out the types and all the boilerplate/inheritance that included ... I like the api, but I'm thinking a single change class might be cleaner ... but I can try and refactor that later, maybe it gets ugly

lib/kennel/syncer.rb
zdrve: Changed to check type `== :update` for now, which is at least much better than using `respond_to?`

PR #250, opened 2023-01-09
Simplify add_tracking_id handling

zdrve:
Before: add tracking ID to all "expected", then selectively remove it again by picking through the update-diffs

Now: only add the tracking ID where we want it.


grosser (approved):
hehe yeah that looks much simpler :D

PR #252, opened 2023-01-09
More functional

zdrve:
Less passing of data via instance variables.

The name `internal_plan` is a bit clunky, but the obvious name `plan` is taken, repeatedly:

 * the "internal" plan (formerly `@create, @update, @delete`)
 * the "external" plan, as returned by ...
 * ... the method `#plan`

Just for bonus giggles, `plan` is of course both a noun and a verb.

We could simplify things a lot by making the external and internal plans the same, and then the name "plan" would work for everything.

### DIFF
--- a/lib/kennel/syncer.rb
+++ b/lib/kennel/syncer.rb
@@ -6,6 +6,7 @@ module Kennel
     LINE_UP = "\e[1A\033[K" # go up and clear
 
     Plan = Struct.new(:changes, keyword_init: true)
+    InternalPlan = Struct.new(:creates, :updates, :deletes)
     Change = Struct.new(:type, :api_resource, :tracking_id, :id)
 
     def initialize(api, expected, actual, strict_imports: true, project_filter: nil, tracking_id_filter: nil)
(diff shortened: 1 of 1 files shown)


lib/kennel/syncer.rb
```diff
@@ -164,14 +167,14 @@ def initialize
         @attribute_differ = AttributeDiffer.new
       end
 
-      def display(create, update, delete)
+      def display(internal_plan)
         Kennel.out.puts "Plan:"
-        if create.empty? && update.empty? && delete.empty?
+        if internal_plan.values.all?(&:empty?)
```
grosser, line 172:
```suggestion
        if noop?
```

zdrve: It's in a different class now 😅

grosser: then it should have the `noop?` method ?

grosser (approved):
the state reduction is nice 👍

zdrve:
Yeah the core is getting a lot easier to understand now!

PR #254, opened 2023-01-10
Prefer tracking id match over datadog id

zdrve:
* bug fix: check the api_resource of an "expected" found via tracking id (e.g. don't match a monitor with a dashboard)
* not necessarily a bug fix: prefer tracking id over datadog id

Regarding the second one: until recently I had always thought that Kennel did in fact work exactly like this. It made it easy for me to reason about what `id:` (i.e. the import feature) was for, and how it behaved. And then I found out that in fact it didn't do that at all. But maybe it should. wdyt?

### DIFF
--- a/lib/kennel/syncer/matched_expected.rb
+++ b/lib/kennel/syncer/matched_expected.rb
@@ -36,7 +36,12 @@ module Kennel
 
         def matching_expected(a, map)
           klass = a.fetch(:klass)
-          map["#{klass.api_resource}:#{a.fetch(:id)}"] || map[a.fetch(:tracking_id)]
+
+          if (e = map[a.fetch(:tracking_id)])
+            return e if e.class.api_resource == klass.api_resource
+          end
+
+          map["#{klass.api_resource}:#{a.fetch(:id)}"]
         end
       end
     end
--- a/test/kennel/syncer/matched_expected_test.rb
+++ b/test/kennel/syncer/matched_expected_test.rb
@@ -129,4 +129,30 @@ describe Kennel::Syncer::MatchedExpected do
       result
     end
   end
+
+  describe "resolution order" do
+    it "prefers tracking_id over id" do
+      e0 = make_expected("a:a", monitor, 999)
+      e1 = make_expected("b:b", monitor, 777)
+      a = make_actual("a:a", monitor, 777)
+      expected << e0
+      expected << e1
+      actual << a
+
+      matched.must_equal [[e0, a]]
+      unmatched_expected.must_equal [e1]
+      unmatched_actual.must_be_empty
+    end
+  end
+
+  it "refuses to match on tracking_id if the api_resource is different" do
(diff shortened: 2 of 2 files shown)


lib/kennel/syncer/matched_expected.rb
```diff
@@ -36,7 +36,12 @@ def matching_expected_lookup_map(expected)
 
         def matching_expected(a, map)
           klass = a.fetch(:klass)
-          map["#{klass.api_resource}:#{a.fetch(:id)}"] || map[a.fetch(:tracking_id)]
+
+          if (e = map[a.fetch(:tracking_id)])
+            return e if e.class.api_resource == klass.api_resource
```
grosser, line 41: can we raise on else ... because that seems broken ...

grosser (commented):
id is added when trying to do funky stuff ... so it should be preferred during matching

zdrve:
> id is added when trying to do funky stuff

Hmm, well that's interesting. What is this "funky stuff" use case?

grosser:
attaching existing monitors or deleting existing ensuring things stay connected correctly when experimenting (for example I removed a tracking id to see if an update that failed would work when creating)

zdrve:
I'm not sure I understand, and I'd like to – this is what I was thinking when I wrote https://zendesk.slack.com/archives/C81T33XTK/p1672864770100369 . Maybe we could write these down somewhere in more detail?

Also because it's related to the "previous tracking id" feature idea – I want to be sure that that feature (a) doesn't conflict with existing use cases and (b) isn't made redundant (by such things already being perfectly possible).

grosser:
the id prevents accidental deletion by "renaming the kennel id" (which can happen by accident when moving project or changing the name which then generates the kennel_id)

idk if it conflicts with "previous tracking id", prefer to POC that feature and see if that's the case and if not then no change to touch existing behavior

zdrve:
Interesting – and this is the first I've heard of this use case. Which now makes me wonder if it was a really bad idea to remove all those `id` properties. Nobody mentioned this use case when I asked for help here https://zendesk.slack.com/archives/C81T33XTK/p1672864770100369 

So `id` effectively has two uses then.

grosser:
If we had a nice way of not deleti+create when changing a kennel id, then
it would be less needed.
afaik removing ids is still a good idea since they mostly come from imports
that were never cleaned up.

lib/kennel/syncer/matched_expected.rb
zdrve: What would be the benefit of raising?

grosser: I'm hoping it would either make a bad PR fail instantly or show exactly what is broken, since I hope this is from a code bug and not someone copy-pasting tracking ids into different items.

With "self heal" you mean kennel would delete on of the items ?
... that might be alright too, since duplicate ids should fail when generating already

... so either way works 👍

PR #256, opened 2023-01-22
Ensure tracking ids don't undergo filesystem mangling

zdrve:
Be more strict about what we allow in tracking ids, thus enforcing the rules which we kind of already take on trust as being true.

### DIFF
--- a/lib/kennel/parts_serializer.rb
+++ b/lib/kennel/parts_serializer.rb
@@ -45,6 +45,7 @@ module Kennel
     end
 
     def path_for_tracking_id(tracking_id)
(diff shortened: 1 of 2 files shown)


grosser (commented):
we already have that in record.rb:113 ?

zdrve:
> we already have that in record.rb:113 ?

Well yes, but you can override it. But if we _do_ trust that that provides the same protection, then we can simplify the mangling to become just `generated/#{tracking_id.sub(":", "/")}.json"`

grosser:
prefer that
alternatively to move the protection check into the mangling, but afaik it has worked well enough so far, so don't see a need to change it

PR #257, opened 2023-01-22
Pass filter object

zdrve:
Always pass the filter around as the filter object, not `project_filter` and `tracking_id_filter`.


grosser (approved):
cleans it up nicely! 🎉

PR #258, opened 2023-01-22
More filtering

zdrve:
Add `Filter#matches_project_id?`. Make the instance variables `project_filter, tracking_id_filter` private.

Thus the interface exposed by Filter is essentially 3 predicates: `filtering?`, `matches_project_id?`, `matches_tracking_id?`.

### DIFF
--- a/lib/kennel/filter.rb
+++ b/lib/kennel/filter.rb
@@ -2,8 +2,6 @@
 
 module Kennel
   class Filter
-    attr_reader :project_filter, :tracking_id_filter
-
     def initialize
       # build early so we fail fast on invalid user input
       @tracking_id_filter = build_tracking_id_filter
@@ -22,14 +20,22 @@ module Kennel
       !project_filter.nil?
     end
 
+    def matches_project_id?(project_id)
+      return true unless filtering?
+
+      project_filter.include?(project_id)
+    end
+
     def matches_tracking_id?(tracking_id)
-      return true if project_filter.nil?
+      return true unless filtering?
       return tracking_id_filter.include?(tracking_id) if tracking_id_filter
 
       project_filter.include?(tracking_id.split(":").first)
     end
 
     private
+
+    attr_reader :project_filter, :tracking_id_filter
 
     def build_project_filter
       project_names = ENV["PROJECT"]&.split(",")&.sort&.uniq
--- a/lib/kennel/filter.rb
+++ b/lib/kennel/filter.rb
@@ -21,9 +21,7 @@ module Kennel
     end
 
     def matches_project_id?(project_id)
-      return true unless filtering?
-
-      project_filter.include?(project_id)
+      !filtering? || project_filter.include?(project_id)
     end
 
     def matches_tracking_id?(tracking_id)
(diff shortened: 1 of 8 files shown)


lib/kennel/filter.rb
```diff
@@ -18,15 +16,27 @@ def filter_parts(parts)
       filter_resources(parts, :tracking_id, tracking_id_filter, "resources", "TRACKING_ID")
     end
 
+    def filtering?
```
grosser, line 19: yeah this is nice hiding 👍

lib/kennel/filter.rb
```diff
@@ -18,15 +16,27 @@ def filter_parts(parts)
...
+      !project_filter.nil?
+    end
+
+    def matches_project_id?(project_id)
+      return true unless filtering?
+
+      project_filter.include?(project_id)
```
grosser, line 26: easier to follow + shorter ?
```suggestion
      filtering? && project_filter.include?(project_id)
```

lib/kennel/parts_serializer.rb
```diff
@@ -35,13 +35,20 @@ def write_changed(parts)
...
+
+      if filter.filtering?
+        segment = Kennel::Models::Record::ALLOWED_KENNEL_ID_SEGMENT
+        paths.select! do |path|
+          if (m = path.match(/^generated\/(#{segment})\/(#{segment})\.json$/o))
+            filter.matches_tracking_id?("#{m[1]}:#{m[2]}")
+          elsif (m = path.match(/^generated\/(#{segment})(?:\/|$)/o)) && File.directory?(path)
+            filter.matches_project_id?(m[1])
+          end
+        end
```
grosser, line 48: same ?
```suggestion
        paths.select! do |path|
          tracking_id = path.split("/")[1..2].to_a.join(":")
          filter.matches_tracking_id?(tracking_id)
        end
```

lib/kennel/syncer/resolver.rb
```diff
@@ -20,19 +20,14 @@ def initialize(expected:, filter:)
...
           next unless tracking_id = a.fetch(:tracking_id)
 
           # ignore when deleted from the codebase
           # (when running with filters we cannot see the other resources in the codebase)
           api_resource = a.fetch(:klass).api_resource
-          next if
-            !id_map.get(api_resource, tracking_id) &&
-            (!project_prefixes || tracking_id.start_with?(*project_prefixes)) &&
-            (!filter.tracking_id_filter || filter.tracking_id_filter.include?(tracking_id))
+          next if !id_map.get(api_resource, tracking_id) && filter.matches_tracking_id?(tracking_id)
```
grosser, line 30: nice cleanup :D

lib/kennel/parts_serializer.rb
zdrve: What you wrote doesn't causes any existing tests to fail, but I'm still reluctant, because it means that `tracking_id` ends up being things which aren't (i.e. can't possibly be) tracking IDs. e.g. `foo` or `a space!`.

It also makes me think that we're missing a test somewhere, but I'm not sure what yet. 🤔

zdrve: Accepting this until I can think of a more concrete thing to add. :-)

grosser (approved):

grosser:
1.132.0

PR #259, opened 2023-01-31
support on_missing_data for event monitors

grosser:
- verified rake plan is clean
- verified it can import + change monitor


sakkas-zendesk (approved):

grosser:
downside is that this ignores notify_no_data and no_data_timeframe options because otherwise updates fail :/

grosser:
so I think this needs more thought ... on_missing_data should be calculated from notify_no_data if it's not set

PR #261, opened 2023-02-10
allow users to paste the generated path of an objects to update it wi…

grosser:
…thout manually converting

sometimes things change I already have the full path and then I'd like to avoid manual conversion
```
rake kennel:update_datadog TRACKING_ID=generated/foo/bar.json
```

### DIFF
--- a/lib/kennel/filter.rb
+++ b/lib/kennel/filter.rb
@@ -45,7 +45,13 @@ module Kennel
     end
 
     def build_tracking_id_filter
-      (tracking_id = ENV["TRACKING_ID"]) && tracking_id.split(",").sort.uniq
+      (tracking_id = ENV["TRACKING_ID"]) && tracking_id.split(",").map { |id| tracking_id_path_conversion(id) }.sort.uniq
+    end
+
+    # allow users to paste the generated path of an objects to update it without manually converting
+    def tracking_id_path_conversion(tracking_id)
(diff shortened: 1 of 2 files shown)


lib/kennel/filter.rb
```diff
@@ -45,7 +45,13 @@ def build_project_filter
...
 
     def build_tracking_id_filter
-      (tracking_id = ENV["TRACKING_ID"]) && tracking_id.split(",").sort.uniq
+      (tracking_id = ENV["TRACKING_ID"]) && tracking_id.split(",").map { |id| tracking_id_path_conversion(id) }.sort.uniq
+    end
+
+    # allow users to paste the generated path of an objects to update it without manually converting
+    def tracking_id_path_conversion(tracking_id)
+      return tracking_id unless tracking_id.end_with?(".json")
+      tracking_id.split(/\/|\.json/).last(2).join(":")
```
zdrve, line 54: We could also do

```suggestion
      tracking_id.sub("generated/", "").sub(".json", "").sub("/", ":")
```

maybe. i.e. both the `generated/` and the `.json` are optional.

zdrve (approved):
Could be useful :-)

lib/kennel/filter.rb
grosser: yeah that reads better

PR #263, opened 2023-02-16
import empty hash as single line to satisfy rubocop Layout/SpaceInsid…

grosser:
…eHashLiteralBraces

letting users run into rubocop issues is not fun for them :)

/cc @zdrve


zdrve (approved):
Could we run the importer output through `rubocop -a`, maybe?

grosser:
yeah that could work ... not sure if it's worth the overhead compared to fixing the output whenever we see trouble

PR #264, opened 2023-02-17
generate when updating too

grosser:
I often mess around in a definition and run update until it does what I want
would be nice to always see the generated diff in git and not have the PR fail on no_diff

happens to make update faster by accident because it was not using preload

@zdrve

### DIFF
--- a/lib/kennel/tasks.rb
+++ b/lib/kennel/tasks.rb
@@ -111,6 +111,8 @@ namespace :kennel do
 
   desc "update datadog (scope with PROJECT=name)"
   task update_datadog: :environment do
+    Kennel::Tasks.kennel.preload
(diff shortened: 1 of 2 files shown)


zdrve (approved):
Yeah, honestly, I'm thinking of rewriting the generate / plan / update_datadog tasks into a single entrypoint in `Kennel::Engine`. (update: #265)

Sketch:
```
Kennel::Engine.new(
  write_generated: true,
  show_plan: true,
  wait_for_confirmation: true,
  update_datadog: true
).run
```

where the different options could be used to implement what we call today `generate`, `plan`, and `update_datadog`.

The fact that these tasks use different entry points into the Engine is what makes the preload thing so hacky.

PR #265, opened 2023-02-18
Fully encapsulate the preloading

zdrve:
By combining all the actions into "run", where the constructor options specify _what_ we want to run, then we give the Engine all the information it needs to execute things in the most time-efficient manner.

Also there are more time savings to be made (#266), and this refactor makes that purely an internal matter for the Engine.

Note that `kennel_test` still says `describe ".plan"` and so forth, even though there is no such method any more.

Note: kennel-config `ci_and_notify` will need updating. I haven't looked into that yet.


lib/kennel.rb
```diff
@@ -54,48 +54,88 @@ class << self
   self.err = $stderr
 
   class Engine
-    attr_accessor :strict_imports
+    def initialize(
+      update_generated: true,
+      show_plan: false,
+      require_confirm: nil,
```
grosser, line 60:
```suggestion
      require_confirm: default_require_confirm?,
```

lib/kennel.rb
```diff
@@ -54,48 +54,88 @@ class << self
   self.err = $stderr
 
   class Engine
-    attr_accessor :strict_imports
+    def initialize(
+      update_generated: true,
```
grosser, line 58:
```suggestion
      generate: true,
```

lib/kennel.rb
```diff
@@ -54,48 +54,88 @@ class << self
...
+      show_plan: false,
+      require_confirm: nil,
+      update_datadog: false,
+      strict_imports: true
```
grosser, line 62: feels a bit strange that actions and options are on the same level

lib/kennel.rb
```diff
@@ -54,48 +54,88 @@ class << self
...
+          default_require_confirm?
+        else
+          require_confirm
+        end
+      @update_datadog = update_datadog
+      @strict_imports = strict_imports
+    end
+
+    def run
+      if !update_generated? && !show_plan? && !update_datadog?
```
grosser, line 77: should fail and say "you did not ask for anything" ?

lib/kennel.rb
```diff
@@ -54,48 +54,88 @@ class << self
...
+      if show_plan? || update_datadog?
+        # start generation and download in parallel to make planning faster
+        Utils.parallel([:generated, :definitions]) { |m| send m, plain: true }
+      end
+
+      if update_generated?
+        PartsSerializer.new(filter: filter).write(generated)
+      end
+
+      if show_plan? || update_datadog?
```
grosser, line 91: store into a variable and reuse from above
```suggestion
      if needs_syncing
```

lib/kennel.rb
```diff
@@ -54,48 +54,88 @@ class << self
...
+        syncer # Instantiating the syncer calculates the plan
 
-    def initialize
-      @strict_imports = true
+        syncer.print_plan if show_plan?
+
+        if update_datadog?
+          syncer.update if update_datadog? && (!require_confirm? || syncer.confirm)
```
grosser, line 97:
```suggestion
        if update_datadog?
          syncer.update if !require_confirm? || syncer.confirm
```

lib/kennel.rb
```diff
@@ -54,48 +54,88 @@ class << self
...
-      syncer.print_plan
-      syncer.update if syncer.confirm
+    def require_confirm?
+      @require_confirm
     end
 
-    private
+    def update_datadog?
+      @update_datadog
+    end
```
grosser, line 126: these accessors feel a bit awkward, would prefer the non `?` version

grosser (commented):
yeah the current version is a bit clumsy in that the caller needs to know about the insides

I still liked that the rake tasks call different methods and the engine had a semi-public interface
now everything is in a single method and nobody can add any logic from the outside

lib/kennel.rb
zdrve: The option to run an action. 🤷‍♀️

zdrve:
> nobody can add any logic from the outside

I feel this argument would carry more weight if I knew of at least one user of this who isn't Zendesk. We _know_ that Zendesk doesn't have that use case; I don't know that anyone else does. So this feels a bit like a solution looking for a problem.

On the other hand, the performance, and the desire to have Zendesk kennel run faster, is definitely a real use case.

lib/kennel.rb
zdrve: It's a shame that ruby does make this easier. `attr_reader :foo, :bar?` would be very handy, if it existed.

lib/kennel.rb
zdrve: It could, that's true.

Or it could do nothing (e.g. just `return nil`).

Or it could do something useful (validate that the parts can be built and are valid). Which is what it does.

lib/kennel.rb
zdrve: Can do, in which case I'll also rename `def generated` (e.g. to `def parts`). It's confusing that `generated` is both an in-memory thing, and also an on-filesystem thing.

lib/kennel.rb
zdrve: It changes the behaviour of `.new(require_confirm: nil, ...)`, but I suppose that's OK :-) It wasn't defined either way.

lib/kennel.rb
zdrve: Done. It's probably a taste thing; I don't see the benefit.

lib/kennel.rb
grosser: afaik that's not a very common pattern, usually there are options for how a thing behaves and then methods are called to perform actions
... I get why it's needed though

lib/kennel/predicate_accessors.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Kennel
+  module PredicateAccessors
+    def self.included(into)
+      into.extend(::Kennel::PredicateAccessors::ClassMethods)
+    end
+
+    module ClassMethods
+      def attr_reader(*symbols)
```
grosser, line 10: maybe simpler and not risk of anything going wrong :)
```suggestion
      def attr_reader?(*symbols)
```

grosser: ... or doing this inline
```ruby
[x].each { |s| define_method s }
```

lib/kennel.rb
```diff
@@ -54,56 +55,78 @@ class << self
...
   class Engine
-    attr_accessor :strict_imports
-
-    def initialize
-      @strict_imports = true
+    def initialize(
+      generate: true,
+      show_plan: false,
+      require_confirm: default_require_confirm?,
+      update_datadog: false,
```
grosser, line 62: nit (or maybe for later) simpler keywords could be nice (but maybe also confusing)
```suggestion
      plan: false,
      update: false,
      require_confirm: default_require_confirm?,
```

lib/kennel.rb
```diff
@@ -54,56 +55,78 @@ class << self
...
+      @update_datadog = update_datadog
+      @strict_imports = strict_imports
     end
 
-    # start generation and download in parallel to make planning faster
-    def preload
-      Utils.parallel([:generated, :definitions]) { |m| send m, plain: true }
-    end
+    def run
+      if !generate? && !show_plan? && !update_datadog?
```
grosser, line 73:
```suggestion
      # when no action is requested we just validate that parts can be rendered
      if !generate? && !show_plan? && !update_datadog?
```

lib/kennel.rb
```diff
@@ -54,56 +55,78 @@ class << self
...
+      if needs_syncing
+        syncer # Instantiating the syncer calculates the plan
+
+        syncer.print_plan if show_plan?
+
+        if update_datadog?
+          syncer.update if !require_confirm? || syncer.confirm
+        else
+          syncer.plan # i.e. get & return the already-calculated plan
+        end
```
grosser, line 98:
```suggestion
        end
      else
        nil
```

lib/kennel/tasks.rb
```diff
@@ -8,18 +8,13 @@
 module Kennel
   module Tasks
     class << self
-      def kennel
-        @kennel ||= Kennel::Engine.new
```
grosser, line 12: this going away is kinda nice

grosser (approved):
seems better than before ... does not feel like a clean solution, but neither was the "before" 😞

grosser:
another approach could be to have `run` take `generate: true, plan: true, update: true` options and `new` only takes `strict_imports`

lib/kennel.rb
zdrve: Yeah, confusing.

Does `plan` mean make a plan, or show a plan? Does `update` mean update `generated`, or update Datadog? For the sake of one extra word, I'd much rather clearer names.

grosser:
Yeah it's clearer

lib/kennel/predicate_accessors.rb
zdrve: Renamed, c36c8e1d6e3582ba0cf02ec8673c55dff92b245a

lib/kennel.rb
zdrve: * `Kennel::Engine.new(**some_options).run`
* `Kennel::Engine.new(**some_options).run(**more_options)`

are equally good/bad IMO. Either way will still have the constraint that the only useful pattern is `.new` followed by `#run` (exactly once).

One could go with more of a 1:1 with the rake tasks:

* Kennel::Engine.new(**some_options).generate`
* Kennel::Engine.new(**some_options).plan`
* Kennel::Engine.new(**some_options).update_datadog`

but (a) that only really makes sense insofar as `rake plan` is actually doing a `parts` and `generate` and `syncer` and then returning the syncer plan, i.e. it's dependent on the internals of those tasks. And (b) we already diverge significantly from this with our `ci_and_notify` or whatever it's called.
 
The existing code already allows for:
 * build parts
 * build parts, save to `generated`
 * build parts, show the plan
 * build parts, save to `generated`, show the plan
 * build parts, show the plan, update datadog
 * build parts, save to `generated`, show the plan, update datadog
 * build parts, show the plan, get confirmation, update datadog
 * build parts, save to `generated`, show the plan, get confirmation, update datadog

so this PR just exposes what's already going on more clearly (one could argue).

grosser: yeah, good point

grosser (approved):

PR #266, opened 2023-02-19
Promises

zdrve:
Call the parts serializer (writing to `generated`) in parallel with the downloads, to shave another second or two off the run time.

This also adds `plain: true` progress mode to "Storing" and "Building json", where it was missing before.

And honestly since parallelism is the way to go for performance, we should probably make "plain" mode the _only_ mode, i.e. remove the "spinner" mode. That would also clear up a chunk of code in `Progress`.


kennel.gemspec
```diff
@@ -15,5 +15,6 @@ Gem::Specification.new name, Kennel::VERSION do |s|
   s.add_runtime_dependency "faraday", "~> 1.8"
   s.add_runtime_dependency "hashdiff", "~> 1.0"
   s.add_runtime_dependency "net-http-persistent", "~> 4.0"
+  s.add_runtime_dependency "rosarium", "~> 0.1"
```
zdrve, line 18: Does this need to be added to `Gemfile` too? I've never understood why there's both a `Gemfile` and a gemspec, both specifying the dependencies.

grosser: gemspec is for what the gem is dependent on
Gemfile is for what the development environment is dependent on

lib/kennel.rb
```diff
@@ -125,6 +135,20 @@ def api
       @api ||= Api.new
     end
 
+    def p_parts
+      @p_parts ||= Rosarium::Promise.execute { parts(plain: true) }
+    end
+
+    def p_definitions
+      @p_definitions ||= Rosarium::Promise.execute { definitions(plain: true) }
+    end
```
grosser, line 144: idk how this is supposed to be faster, old code does them in serial ... and we can't really do them in parallel right ?
```
      parts = generated
      PartsSerializer.new(filter: filter).write(parts) if ENV["STORE"] != "false" # quicker when debugging
      parts
```

zdrve: The faster part (only by a second or two) is that the `PartsSerializer` doesn't need to wait for the `definitions` to be ready. So writing to `generated` can happen in parallel with downloading from the API.

kennel.gemspec
zdrve: TIL :-)

lib/kennel.rb
grosser: ah PartsSerializer did not run in the thread ...

but so all we need is:
```
todo = []
todo.push :parts, :definitions if needs_syncing
todo.push :serialize_parts if generate?
Utils.parallel(todo) { |m| send m, plain: true }
```

grosser: corrected the code

zdrve: That would be a minimal way of achieving the desired change. I think it might be a bit short-sighted though – but without me pointing to _further_ benefits in this direction, "short-sighted" may be just fine :-)

PR #268, opened 2023-02-23
Add support for scheduling_options on monitors

dannytranlx:
Hi @grosser, I took a quick swing at adding support for `scheduling_options` on monitors. This would help with log quota related monitors resetting on a specific time:

image

image

Implementation isn't exhaustive, but would allow users to pass an object in to set their schedule. Feel free to edit if needed.

Thank you!

### DIFF
--- a/lib/kennel/models/monitor.rb
+++ b/lib/kennel/models/monitor.rb
@@ -32,7 +32,7 @@ module Kennel
       settings(
         :query, :name, :message, :escalation_message, :critical, :type, :renotify_interval, :warning, :timeout_h, :evaluation_delay,
         :ok, :no_data_timeframe, :notify_no_data, :notify_audit, :tags, :critical_recovery, :warning_recovery, :require_full_window,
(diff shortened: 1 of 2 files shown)


EricHogueZD (approved):

dygcao (approved):

grosser:
- generate shows no diff 👍 
- update shows no diff 👍 
- was unable to see this in the UI to set/unset it ... how did you get that UI ?

dannytranlx:
It's hidden in the evaluation window dropdown: 
current_day

grosser:
verified that removing works

adding I ran into:
```
{"errors":["Monitors with an evaluation_window scheduling_option must have a current query period and vice versa."]}
```
... but as long as a clean import works or you had it working at all then that's all good
lmk if you want to merge+release now

dannytranlx:
Tests are good on my end. Clean import with `current_1d` and update existing monitors to `current_1d`.

I was able to reproduce the same error you mentioned, it happens when the query has `current_1d`, but without the `scheduling_option` field. The error is expected if the field is missing.

It's good on my end if you'd like to merge and release, thank you!

PR #273, opened 2023-03-13
Add `Kennel.in`; test tidying

zdrve:
- Add `Kennel.in`
- In `capture_all`, set up `Kennel.in` to something consistent (so we don't rely on the test execution environment)
- Use `Kennel.in/out/err` consistently (never use `STD...` nor `$std...`)
- Give `reset_instance` a better name
- Remove redundant `Kennel.@instance` reset

## Checklist
- [ ] Added tests


grosser (approved):
nice consistency ❤️

PR #280, opened 2023-05-26
enforce sensible tags + lowercase for slo

grosser:
cleanup PR should be merged before bumping gem version


lib/kennel/tags_validation.rb
```diff
@@ -0,0 +1,15 @@
...
+module Kennel
+  module TagsValidation
+    def validate_json(data)
+      super
+
+      # ideally we'd avoid duplicated tags, but that happens regularly when importing existing monitors
+      data[:tags] = data[:tags].uniq
+
+      # keep tags clean (TODO: reduce this list)
+      bad_tags = data[:tags].grep(/[^A-Za-z:_0-9.\/*@!#-]/)
```
zdrve, line 11: Why that set of characters?

grosser: set that we currently need ... ideally it would be less

PR #281, opened 2023-05-26
allow ignoring unignored errors

grosser:
monitors that are meant to be reused can ignore builtin errors
but when someone reuses them, they should not have to unignore the ignored errors
(ideally they would, but dealing with 200+ legacy usages that I don't want to go back and fix + this is a neat feature when doing big refactors)

also refactoring to:
- to not call methods that often
- distinguish between errors and tags

### DIFF
--- a/lib/kennel/optional_validations.rb
+++ b/lib/kennel/optional_validations.rb
@@ -4,6 +4,7 @@ module Kennel
     ValidationMessage = Struct.new(:tag, :text)
 
     UNIGNORABLE = :unignorable
+    UNUSED_IGNORES = :unused_ignores
 
     def self.included(base)
       base.settings :ignored_errors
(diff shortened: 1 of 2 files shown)


zdrve:
1. I don't understand what this PR is trying to achieve, neither from the title nor the PR description
2. If there's refactoring involved, maybe do that in a separate PR for clarity. Or at least separate commits.

grosser:
it allows ignoring the "there are ignored errors that did not need to be
ignored" error

zdrve:
I must have my slow head on.

What's the use case here? i.e. what problem are we trying to solve?

Does this need to be invoked in a particular way, or does this change the behaviour of an existing case?

grosser:
we have monitors with ignored errors that don't have these errors
but removing the ignores does not work because they are heavily reused
so we need a way to ignore having "useless ignored errors"

grosser:
refactor extract https://github.com/grosser/kennel/pull/282

grosser:
rebased on refactor PR

PR #284, opened 2023-05-27
do not throw/catch during error building, just let it fail

grosser:
if validation crashes then just let it ... user has to fix it and gets a clean backtrace to do it
instead of us hiding the backtrace and adding more complexity

@zdrve

### DIFF
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -2,12 +2,6 @@
 module Kennel
   module Models
     class Record < Base
-      class PrepareError < StandardError
-        def initialize(tracking_id)
-          super("Error while preparing #{tracking_id}")
-        end
-      end
-
       include OptionalValidations
 
       # Apart from if you just don't like the default for some reason,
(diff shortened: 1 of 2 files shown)


lib/kennel/models/record.rb
```diff
@@ -142,18 +136,10 @@ def build_json
...
-          json = build_json
-          (id = json.delete(:id)) && json[:id] = id
-          validate_json(json)
-        rescue StandardError
-          if unfiltered_validation_errors.empty?
-            @unfiltered_validation_errors = nil
-            raise PrepareError, safe_tracking_id # FIXME: this makes errors hard to debug when running tests
-          end
-        end
+
```
zdrve, line 139: I'm not convinced that this is a good idea.

In case it wasn't clear, the purpose of this code is as follows: validation errors _should_ nicely raise an exception via `invalid!`; the build / validation code shouldn't raise other exceptions. This is what allows us to report _all_ of the problems to the user, rather than trickle-feeding one problem at a time.

However: the purpose of this code is this: if a record has already raised an error via `invalid!`, then we "forgive" any subsequent crashes from the validation code for this item, since the item is obviously already known to be in an untrusted state.

test/kennel/models/record_test.rb
```diff
@@ -55,42 +55,6 @@ def self.parse_url(*)
       record.as_json.must_equal some_json
     end
 
-    it "throws if build_json throws, and there were no validation errors" do
```
zdrve, line 58: Since we're mostly only _removing_ code in this PR, and the purpose of the PR is to make backtraces more visible, where is the test which proves that we've achieved that goal?

grosser: I usually don't add tests for behavior that does not exist, i.e. "it does not suppress errors" since that would be then in theory be needed for every method
... I'll verify that backtraces look sane though

lib/kennel/models/record.rb
grosser: my thinking was:
there should not be exceptions in the validation code because that means the library made  mistake -> fix the library instead of hiding the problem (and making it harder to fix)

grosser: ... and if exceptions are expected then any exception should be captured and not just "exceptions after validation errors" so it's consistent

test/kennel/models/record_test.rb
grosser: yeah it raises cleanly as expected

zdrve (approved):

PR #285, opened 2023-05-27
simplify as_json

grosser:
basically fix all tests so we don't need this crutch anymore

### DIFF
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -8,8 +8,6 @@ module Kennel
         end
       end
 
-      UnvalidatedRecordError = Class.new(StandardError)
-
       include OptionalValidations
 
       # Apart from if you just don't like the default for some reason,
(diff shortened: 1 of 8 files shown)


zdrve (approved):
Yeah, this is why I made https://github.com/grosser/kennel/pull/234 – because I a mechanism whereby `as_json` simply doesn't exist unless you've gone through the right call flow. I still think that would be a better way to go, tbh.

That said, there's nothing wrong with this PR, other than that it removes a layer of protection, which _maybe_ we don't need.

grosser:
prefer not to have the layer so we find places where things are called out of order
... it's a bit annoying to always build, but at least it's always the same flow :)

zdrve:
> prefer not to have the layer so we find places where things are called out of order

If you mean what I think you mean, then you'd still find places where things are called out of order, so this is an invalid argument.

> it's a bit annoying to always build

It's how one ensures that the checks have been applied :-)

PR #286, opened 2023-05-28
extract optional validation

grosser:
filter + ignore responsibility should live in optional_validations, not inside the record
so extracting calls and making the methods static
results in 1 less state being kept on records

~includes https://github.com/grosser/kennel/pull/284~ (edit: I changed the base branch)

/cc @zdrve

### DIFF
--- a/lib/kennel/models/record.rb
+++ b/lib/kennel/models/record.rb
@@ -77,7 +77,7 @@ module Kennel
         end
       end
 
-      attr_reader :project, :unfiltered_validation_errors, :filtered_validation_errors, :as_json
+      attr_reader :project, :validation_errors, :as_json
 
       def initialize(project, *args)
         raise ArgumentError, "First argument must be a project, not #{project.class}" unless project.is_a?(Project)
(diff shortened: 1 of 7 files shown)


lib/kennel/optional_validations.rb
```diff
@@ -11,17 +11,17 @@ def self.included(base)
     end
 
     def self.valid?(parts)
-      parts_with_errors = parts.reject do |part|
-        part.filtered_validation_errors.empty?
+      parts_with_errors = parts.filter_map do |part|
```
zdrve, line 14: Why not just `map`?

lib/kennel/models/record.rb
```diff
@@ -77,7 +77,7 @@ def ignore_default(expected, actual, defaults)
         end
       end
 
-      attr_reader :project, :unfiltered_validation_errors, :filtered_validation_errors, :as_json
+      attr_reader :project, :unfiltered_validation_errors, :as_json
```
zdrve, line 80: If the Record doesn't really know about filtering, then why is this called "unfiltered" validation errors?

lib/kennel/optional_validations.rb
grosser: fixed

lib/kennel/models/record.rb
grosser: renamed to validation_errors

PR #287, opened 2023-05-30
move all validation related fields/methods to OptionalValidations

grosser:
optional validations needs errors and ignored errors ... but somehow it only took care of half of them
so either they should move all into OptionalValidations (this PR)
or all move into record 🤷 
/cc @zdrve


zdrve (approved):
Didn't read too closely but it looks believable :-)

PR #288, opened 2023-06-03
validate warning and critical match each other

grosser:
/cc @zdrve


zdrve (approved):
All good apart from the failing test

PR #298, opened 2023-09-28
Remove renotify interval restriction

petern-sc:
Datadog allows arbitrary intervals now so no need to restrict the interval values

image

---------------------------

image

-------------------

image

## Checklist
- [ ] Added tests


grosser:
1.144.2

PR #303, opened 2023-11-02
fix symbol validation not working in dashboards

grosser:
## Checklist
- [X] Verified against local install of kennel (using `path:` in Gemfile)
- [X] Added tests


lib/kennel/models/record.rb
```diff
@@ -160,7 +160,7 @@ def validate_json(data)
         bad = Kennel::Utils.all_keys(data).grep_v(Symbol).sort.uniq
         return if bad.empty?
         invalid!(
-          :hash_keys_must_be_symbols,
+          OptionalValidations::UNIGNORABLE,
```
zdrve, line 163: Nice! I forgot it would be so simple! 😅

PR #304, opened 2023-11-02
ensure diffing is always fast by being strict about int vs float

grosser:
## Checklist
- [X] Verified against local install of kennel (using `path:` in Gemfile)


grosser:
ruby `==` already does `1 == 1.0` so we can't add this and don't need to add this

PR #305, opened 2023-12-14
Expand on the help for the mentions check

ghost:
In particular, suggest rebasing

**So far, untested**

Also, maybe the fix for this should be in zendesk-kennel as well or instead of here.

### DIFF
--- a/lib/kennel/tasks.rb
+++ b/lib/kennel/tasks.rb
@@ -110,9 +110,11 @@ namespace :kennel do
         - If the problems listed below seem valid, i.e. a typo or other mistake,
           then fix them :-)
 
-        - If you want to mention someone new, someone whom no existing monitor
-          currently mentions, then you can allow-list them by adding to the
-          `KNOWN` environment variable, or by adding them via #{url}
+        - If you want to mention someone whom no existing monitor
+          currently mentions, then add them via #{url}
+
+        - If you cannot add to above url, ensure the mention actually works by sending a test message, and
+          if it does, add the mention to the `KNOWN` environment variable
 
       MESSAGE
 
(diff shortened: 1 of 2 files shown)


lib/kennel/tasks.rb
```diff
@@ -91,7 +91,31 @@ def git_push?
 
     if bad.any?
       url = Kennel::Utils.path_to_url "/account/settings"
-      Kennel.out.puts "Invalid mentions found, either ignore them by adding to `KNOWN` env var or add them via #{url}"
+
+      Kennel.out.puts <<~MESSAGE
+        ⛔️ Possibly invalid mentions found ⛔️
+
+        The monitor definitions in `generated` include at least one mention
```
grosser, line 98:
```suggestion
        The monitor definitions in `generated/` include at least one mention
```

lib/kennel/tasks.rb
```diff
@@ -91,7 +91,31 @@ def git_push?
...
+
+        - If the problems listed below are nothing to do with what you're
+          changing on this branch, then rebase onto a fresh `#{ENV.fetch("DEFAULT_BRANCH", "master")}`
+
+        - If the problems listed below seem valid, i.e. a typo or other mistake,
+          then fix them :-)
+
+        - If you want to mention someone new, someone whom no existing monitor
+          currently mentions, then you can allow-list them by adding to the
+          `KNOWN` environment variable, or by adding them via #{url}
```
grosser, line 115:
```suggestion
        - If you want to mention someone whom no existing monitor
          currently mentions, then add them via #{url}

        - If you cannot add to above url, ensure the mention actually works by sending a test message and
          if it does add to `KNOWN` environment variable
```

grosser (commented):
much better :)

grosser (approved):

zdrve:
btw the reason I haven't merged this is that while this PR I think is OK as a standalone thing, I haven't yet got a good understanding of how this would in conjunction with the equivalent code in zendesk-kennel.

PR #307, opened 2024-02-13
Send all invalid mentions output to stderr

zdrve:
To ensure the "validate mentions" diagnostics come out in an understandable sequence.

It currently ([example](https://github.com/zendesk/kennel/actions/runs/7882274695/job/21507208628?pr=22031#step:5:23)) says:

```text
It generally means that you're trying to add a notification to somewhere for the first time
... (many lines later) ...
Invalid mentions found
```

which makes no sense. So change it to:

```text
Invalid mentions found
...
It generally means that you're trying to add a notification to somewhere for the first time
```

### DIFF
--- a/lib/kennel/tasks.rb
+++ b/lib/kennel/tasks.rb
@@ -91,8 +91,8 @@ namespace :kennel do
 
     if bad.any?
       url = Kennel::Utils.path_to_url "/account/settings"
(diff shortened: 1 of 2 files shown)


grosser (approved):
✨

grosser:
v1.148.1

PR #309, opened 2024-03-02
Log / respect rate limits

zdrve:

## Checklist
- [x] Added tests

https://github.com/zendesk/kennel/actions/runs/8120685832/job/22198172379#step:7:27

### DIFF
--- a/lib/kennel/api.rb
+++ b/lib/kennel/api.rb
@@ -5,6 +5,8 @@ module Kennel
   class Api
     CACHE_FILE = ENV.fetch("KENNEL_API_CACHE_FILE", "tmp/cache/details")
 
+    RateLimitParams = Data.define(:limit, :period, :remaining, :reset, :name)
+
     def self.tag(api_resource, reply)
       klass = Models::Record.api_resource_map[api_resource]
       return reply unless klass # do not blow up on unknown models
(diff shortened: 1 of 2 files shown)


lib/kennel/api.rb
```diff
@@ -5,6 +5,8 @@ module Kennel
   class Api
     CACHE_FILE = ENV.fetch("KENNEL_API_CACHE_FILE", "tmp/cache/details")
 
+    RateLimitParams = Data.define(:limit, :period, :remaining, :reset, :name)
```
grosser, line 8: feels like you had too much golang for breakfast :D

grosser (approved):
nice 🎉

grosser:
good to just merge right ... any gotchas you expect ?
... even if we do this in parallel it should be mostly fine right ?

zdrve:
> good to just merge right ... any gotchas you expect ? ... even if we do this in parallel it should be mostly fine right ?

I think it's good, I've tested it out a fair bit (e.g. https://github.com/zendesk/kennel/actions/runs/8121811687 and others). 

If the rate limit hits, then each thread that hits the rate limit will sleep – so typically you'd suddenly see 10 messages (1 per thread) saying "Rate limited, sleeping for [e.g.] 25 seconds". Then they all wake up and try again.

grosser:
v1.149.0

PR #310, opened 2024-03-02
Wrap update details in a collapsible group

zdrve:
In the ci-and-update output, the details of every changed field, especially for dashboards, can be very large – and that's especially annoying if it's just noise, because they're not your dashboards.

Wrap the update fields in a `::group::` so that they can be collapsed.

billede

https://github.com/zendesk/kennel/actions/runs/8121982186/job/22200952160?pr=22287#step:7:2334

## Checklist
- [ ] Added tests

### DIFF
--- a/lib/kennel/syncer/plan_printer.rb
+++ b/lib/kennel/syncer/plan_printer.rb
@@ -22,11 +22,19 @@ module Kennel
 
       def print_changes(step, list, color)
         return if list.empty?
+
+        use_groups = ENV.key?("GITHUB_STEP_SUMMARY")
+
         list.each do |item|
+          # No trailing newline
+          Kennel.out.print "::group::" if item.class::TYPE == :update && use_groups
+
           Kennel.out.puts Console.color(color, "#{step} #{item.api_resource} #{item.tracking_id}")
           if item.class::TYPE == :update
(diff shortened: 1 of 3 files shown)


lib/kennel/syncer/plan_printer.rb
```diff
@@ -22,11 +22,19 @@ def print(plan)
 
       def print_changes(step, list, color)
         return if list.empty?
+
+        use_groups = ENV.key?("GITHUB_STEP_SUMMARY")
```
grosser, line 26:
```suggestion
        use_groups = ENV.key?("GITHUB_STEP_SUMMARY") && item.class::TYPE == :update
```

test/kennel/syncer_test.rb
```diff
@@ -239,6 +240,19 @@ def change(*args)
       TEXT
     end
 
+    it "wraps in a group, if running under GitHub" do
```
grosser, line 243:
```suggestion
    it "wraps in a group to prevent big dumps like dashboard from making things unreadable, if running under GitHub" do
```

grosser (approved):
nice :D

grosser:
v1.149.0

PR #311, opened 2024-03-29
Remove locked attribute

brianburnszd:
According to Datadog docs, [`locked` attribute in monitors is no longer supported.](https://docs.datadoghq.com/monitors/guide/how-to-set-up-rbac-for-monitors/?tab=ui#api)

This PR:
1. Stops sending `locked` as an attribute when creating monitors
1. No longer diffs `locked` attribute when comparing monitors against Datadog

Tested out locally and shows the API request to create monitors without locked attr:
```
{
  "name": "Test Locked attr",
  "tags": ...
  "options": {
    ...
    <no "locked" option>
    }
}
```

Note that when merged, all existing monitor generated JSON will also loose the `locked` attribute, but because the diff is a no-op, Datadog will not be updated (what we want).

## Checklist
- [ ] Added tests

### DIFF
--- a/test/kennel/models/monitor_test.rb
+++ b/test/kennel/models/monitor_test.rb
@@ -41,7 +41,6 @@ describe Kennel::Models::Monitor do
         include_tags: true,
         escalation_message: nil,
         evaluation_delay: nil,
-        locked: false,
         renotify_interval: 0,
         thresholds: { critical: 123.0 },
         variables: nil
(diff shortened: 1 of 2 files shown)


test/kennel/models/monitor_test.rb
```diff
@@ -581,6 +580,13 @@ def diff_resource(e, a)
       diff_resource({ type: -> { "query alert" } }, {}).must_equal []
     end
 
+    describe "#locked" do
+      it "ignores deprecated option locked" do
+        expected_basic_json[:options].delete(:locked)
```
grosser, line 585: seems sus ... it's already removed from expected_basic_json above ?

brianburnszd: added a commit to address this 👍

brianburnszd:
This may be a breaking change to anyone using `locked: true` in monitor configuration. Not sure how needs to be communicated or if it needs to be noted in a Changelog? Probably requires a major version update?

grosser:
I'm trying a different approach with putting it into read-only, I'll report
back in a bit

grosser:
readonly did not work since it's a option and not a top-level attribute

grosser:
I'll add another small fix and then make zendesk PR

PR #313, opened 2024-05-23
Add support for time_slice slo

kaarolch:
Add suport for `time_slice` [SLO](https://docs.datadoghq.com/service_management/service_level_objectives/time_slice/), that instead `query` use `sliSpecification`: 
```
sli_specification {
    time_slice {
      query {
        formula {
          formula_expression = "query1"
        }
        query {
          metric_query {
            name  = "query1"
            query = "avg:my.custom.count.metric{*}.as_count()"
          }
        }
      }
      comparator = ">"
      threshold  = 0.9
    }
``` 

I'm a beginner at Ruby, so if some of my code isn't well-written, I would appreciate feedback.

## Checklist
- [x] Added tests

### DIFF
--- a/lib/kennel/importer.rb
+++ b/lib/kennel/importer.rb
@@ -3,7 +3,7 @@
 module Kennel
   class Importer
     TITLES = [:name, :title].freeze
-    SORT_ORDER = [*TITLES, :id, :kennel_id, :type, :tags, :query, *Models::Record.subclasses.map { |k| k::TRACKING_FIELDS }, :template_variables].freeze
+    SORT_ORDER = [*TITLES, :id, :kennel_id, :type, :tags, :query, :sli_specification, *Models::Record.subclasses.map { |k| k::TRACKING_FIELDS }, :template_variables].freeze
 
     def initialize(api)
       @api = api
@@ -19,9 +19,14 @@ module Kennel
         raise(ArgumentError, "#{resource} is not supported")
 
       data = @api.show(model.api_resource, id)
+      # get sli specification before normalization.
+      sli_specification = data[:sli_specification] || data.dig(:sli_specification, :time_slice)
+
       id = data.fetch(:id) # keep native value
       model.normalize({}, data) # removes id
       data[:id] = id
+
+      data[:sli_specification] = sli_specification if sli_specification
 
       title_field = TITLES.detect { |f| data[f] }
       title = data.fetch(title_field)
--- a/lib/kennel/models/slo.rb
+++ b/lib/kennel/models/slo.rb
@@ -35,7 +35,7 @@ module Kennel
           type: type
         )
 
-        if v = sliSpecification
+        if type == 'time_slice' && v = sliSpecification
           data[:sliSpecification] = v
         elsif v = query
           data[:query] = v
--- a/lib/kennel/models/slo.rb
+++ b/lib/kennel/models/slo.rb
@@ -14,7 +14,7 @@ module Kennel
         thresholds: []
       }.freeze
 
-      settings :type, :description, :thresholds, :query, :tags, :monitor_ids, :monitor_tags, :name, :groups, :sliSpecification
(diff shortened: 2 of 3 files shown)


lib/kennel/models/slo.rb
```diff
@@ -35,7 +35,9 @@ def build_json
           type: type
         )
 
-        if v = query
+        if type == 'time_slice' && v = sliSpecification
+          data[:sliSpecification] = v
```
grosser, line 39: does it only support sliSpecification or is setting query also fine ?
... because if it needs sliSpecification we should do this:
```suggestion
        if type == 'time_slice'
          data[:sliSpecification] = sli_specification
```

grosser (commented):
the importer needs to support it too

test/kennel/models/slo_test.rb
```diff
@@ -11,12 +11,17 @@ class TestSlo < Kennel::Models::Slo
 
   def slo(options = {})
     Kennel::Models::Slo.new(
+      # Determine the type from options or default to "metric"
```
grosser, line 14: longer to read than the code itself 🤷 
```suggestion
```

test/kennel/models/slo_test.rb
```diff
@@ -11,12 +11,17 @@ class TestSlo < Kennel::Models::Slo
...
       options.delete(:project) || project,
-      {
-        type: -> { "metric" },
+      slo_options ={
+        type: -> { type },
         name: -> { "Foo" },
         kennel_id: -> { "m1" }
-      }.merge(options)
+      }
+      # Conditionally include sliSpecification if type is "time_slice"
```
grosser, line 22:
```suggestion
```

test/kennel/models/slo_test.rb
```diff
@@ -11,12 +11,17 @@ class TestSlo < Kennel::Models::Slo
...
+      slo_options[:sliSpecification] = -> { options[:sliSpecification] } if type == "time_slice"
```
grosser, line 23: idk why this is needed, we can always set it 🤷

grosser:
looks mostly right, thx for the PR
I can pair on this next week if you are getting stuck or spending >30m on this 👍

test/kennel/models/slo_test.rb
kaarolch: Without this statement I can not generate SLO for time series, I return me error that: 

```
rake aborted!
ArgumentError: 'sliSpecification' on Kennel::Models::Slo was not set or passed as option for project knowledge-api on .../knowledge_api/knowledge_api.rb:146:in `new'
```

grosser: then the call in 88 needs to be `sliSpecification: -> { sli_spec }`

kaarolch:
I will try to update imported this week / Mon. Thank for review and hints 👍

lib/kennel/importer.rb
```diff
@@ -19,10 +19,15 @@ def import(resource, id)
         raise(ArgumentError, "#{resource} is not supported")
 
       data = @api.show(model.api_resource, id)
+      # get sli specification before normalization.
+      sli_specification = data[:sli_specification] || data.dig(:sli_specification, :time_slice)
```
grosser, line 23: how can it be in 2 different places ?

lib/kennel/models/slo.rb
```diff
@@ -35,7 +35,9 @@ def build_json
           type: type
         )
 
-        if v = query
+        if type == 'time_slice'
+          data[:sliSpecification] = :sli_specification
```
grosser, line 39: this needs to be the actual call ?
```suggestion
          data[:sliSpecification] = sli_specification
```

lib/kennel/importer.rb
grosser: ... fixed this, we had some temp-fix code that added it to read-only :D

grosser:
fixing here https://github.com/grosser/kennel/pull/314

PR #314, opened 2024-05-25
Kaarolch kch/slo time slice

grosser:
- imported + updated + no diff 🎉 
- `rake generate` not producing tons of diff
- `rake plan`  not producing tons of diff

SCR-20240525-ait

### DIFF
--- a/lib/kennel/importer.rb
+++ b/lib/kennel/importer.rb
@@ -3,7 +3,7 @@
 module Kennel
   class Importer
     TITLES = [:name, :title].freeze
-    SORT_ORDER = [*TITLES, :id, :kennel_id, :type, :tags, :query, *Models::Record.subclasses.map { |k| k::TRACKING_FIELDS }, :template_variables].freeze
+    SORT_ORDER = [*TITLES, :id, :kennel_id, :type, :tags, :query, :sli_specification, *Models::Record.subclasses.map { |k| k::TRACKING_FIELDS }, :template_variables].freeze
 
     def initialize(api)
       @api = api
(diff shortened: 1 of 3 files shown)


grosser:
@kaarolch lmk if there are any gotchas missing ... seems to mostly work from my limited testing with 1 new timeslice slo🤞

grosser:
I had to change the test too because the example json was not correct (was not what the importer spat out for my example)

lib/kennel/importer.rb
```diff
@@ -2,8 +2,15 @@
 
 module Kennel
   class Importer
+    # title will have the lock symbol we need to remove when re-importing
     TITLES = [:name, :title].freeze
-    SORT_ORDER = [*TITLES, :id, :kennel_id, :type, :tags, :query, *Models::Record.subclasses.map { |k| k::TRACKING_FIELDS }, :template_variables].freeze
```
grosser, line 6: FYI this had an unrealted bug too where we needed flat_map because TRACKING_FIELDS can have multiple entries

lib/kennel/models/slo.rb
```diff
@@ -4,7 +4,10 @@ module Models
     class Slo < Record
       include TagsValidation
 
-      READONLY_ATTRIBUTES = superclass::READONLY_ATTRIBUTES + [:type_id, :monitor_tags, :target_threshold, :timeframe, :warning_threshold]
+      READONLY_ATTRIBUTES = [
+        *superclass::READONLY_ATTRIBUTES,
+        :type_id, :monitor_tags, :target_threshold, :timeframe, :warning_threshold
+      ].freeze
```
grosser, line 10: this missing freeze was hiding the issue with the importer removing sli_specification during normalize (we added sli_specification to READONLY_ATTRIBUTES in a local hack)

kaarolch (approved):

kaarolch:
I've tested it in the local branch and object was correctly imported and also json was generated. :ty

PR #318, opened 2025-03-14
Make timeframe editable for SLOs in kennel

gaya33:

## Checklist
- [ ] Added tests


grosser (commented):
does timeframe maybe need to be in sync with the thresholds or something like that ?
... see https://github.com/grosser/kennel/pull/227 which claims it duplicates the existing attributes

PR #341, opened 2025-10-16
Alternate take - still using explicit `primary`

mbibaud:

## Checklist
- [ ] Added tests

### DIFF
--- a/lib/kennel/importer.rb
+++ b/lib/kennel/importer.rb
@@ -21,6 +21,13 @@ module Kennel
       data = @api.show(model.api_resource, id)
 
       id = data.fetch(:id) # keep native value
+
+      if resource == "slo"
+        # we ignore the 3 copy-pasted fields that track the primary threshold, so we need to sort imported thresholds
+        # or the primary information is lost
+        data[:thresholds].sort! { |t| t[:timeframe] == data[:timeframe] ? 0 : 1 }
+      end
+
       model.normalize({}, data) # removes id
       data[:id] = id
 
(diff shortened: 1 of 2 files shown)


lib/kennel/importer.rb
```diff
@@ -21,6 +21,13 @@ def import(resource, id)
       data = @api.show(model.api_resource, id)
 
       id = data.fetch(:id) # keep native value
+
+      if resource == "slo"
+        # we ignore the 3 copy-pasted fields that track the primary threshold, so we need to sort imported thresholds
+        # or the primary information is lost
+        data[:thresholds].sort! { |t| t[:timeframe] == data[:timeframe] ? 0 : 1 }
```
grosser: this does not work, when creating a slo datadog always picks the lowest one as primary

grosser: ah nvm it should work with the slo change 🤞

lib/kennel/models/slo.rb
```diff
@@ -98,6 +95,9 @@ def self.normalize(expected, actual)
           threshold.delete(:target_display)
         end
 
+        # remove copy-pasted values, if thresholds change these will also change
+        [:warning_threshold, :target_threshold].each { |a| actual[a] = expected[a] }
```
grosser: needs to delete timeframe or the importer will show it

grosser:
```
if resource == "slo"
  data[:primary] = data.delete(:timeframe)
end
```
... at that point let's just name this timeframe 🤷

PR #379, opened 2026-06-03
support using sli_specification

grosser:
the way dd seems to work is that once sli_specification is set you need to keep using it since query is then ignored
also we had non time_slice slos now using sli_specification

image

- show sli_specification when importing to encourage using the new format
- when sli_specification is set only send that
- when query is set only send that ... which is not great because if a slo was migrated to "bad events" then the query will be completely ignored ... but if it was created with just sli_specification everything works fine and afaik there is no way to tell what is what

### DIFF
--- a/lib/kennel/models/project.rb
+++ b/lib/kennel/models/project.rb
@@ -10,10 +10,16 @@ module Kennel
 
       def self.file_location
         return @file_location if defined?(@file_location)
-        if (location = instance_methods(false).first)
-          @file_location = force_relative_path(instance_method(location).source_location.first)
+        methods = instance_methods(false)
+        if methods.any?
+          @file_location = methods.detect do |method|
+            location = instance_method(method).source_location.first
+            if (path = find_relative_path(location))
+              break path
+            end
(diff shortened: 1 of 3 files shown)


lib/kennel/models/project.rb
```diff
@@ -10,10 +10,16 @@ class Project < Base
 
       def self.file_location
         return @file_location if defined?(@file_location)
-        if (location = instance_methods(false).first)
-          @file_location = force_relative_path(instance_method(location).source_location.first)
+        methods = instance_methods(false)
```
grosser, line 13: had to change this because I was getting errors I assume because some method ordering changed,
but also saw these before so I assume this bug was in there already

```
    - -- Managed by kennel zra_ingestor:zra-ingestor-high-retries-exceeded-staging in vendor/bundle/ruby/4.0.0/gems/kennel-2.20.0/lib/kennel/settings_as_methods.rb, do not modify manually
    + -- Managed by kennel zra_ingestor:zra-ingestor-high-retries-exceeded-staging in projects/zra_ingestor/zra_ingestor.rb, do not modify manually
```

